Issue #9, opened 2016-08-18
Gin Framework support


johnzeng:
Do you guys have a plan to support gin-go?
Or, do you have any idea can make it possible to allow me to build one by myself?

bbrodriges:
Well, you can add newrelic to gin fairly easy:
```go
func NewrelicMiddleware(appName string, key string) gin.HandlerFunc {

	if appName == "" || key == "" {
		return func(c *gin.Context) {}
	}

	config := newrelic.NewConfig(appName, key)
	app, err := newrelic.NewApplication(config)

	if err != nil {
                panic(err)
        }

	return func(c *gin.Context) {
		txn := app.StartTransaction(c.Request.URL.Path, c.Writer, c.Request)
		defer txn.End()
		c.Next()
	}
}
```

johnzeng:
@bbrodriges I have tried this but it's not getting the errors from my response, so I think this is not good enough for me. I think a better way is put the txn into the gin.Context so I can make more segment and set error for it.

bbrodriges:
@johnzeng totally agree with you. I have wrote a small wrapper to log errors with new relic transaction in gin context:

```go
// WrapError records error using New Relic transaction
// and returns err back to caller function.
// Example:
//    myint, err := strconv.Atoi("golang")
//    if newrelic.WrapError(c, err) != nil {
//       panic(err)
//    }
func WrapError(c *gin.Context, err error) error {
	if err != nil {
		if txn, ok := c.Get("nrtxn"); ok {
			if nrtxn, ok := txn.(newrelic.Transaction); ok {
				nrtxn.NoticeError(err)
			}
		}
	}

	return err
}
```

willnewrelic:
Hi All,

Proposed Gin support available here:  https://github.com/newrelic/go-agent/pull/28

Give it a try!  We are eager to get your feedback.

PR #11, opened 2016-08-22
Add support for WrapHTTPHandler

polds:
References #9.

Adds support for using `http.ListenAndServe`. Makes an opinionated assumption that the transaction name is the same as the requested URL Path.

A sad side effect is that the injected ResponseWriter can't be asserted (At least in Martini)

EDIT: I'm pretty sure this is due to Martini's dependency injection. 

Please see my test below:

```
package main

import (
    "errors"
    "log"
    "net/http"
    "time"

    "github.com/go-martini/martini"
    "github.com/newrelic/go-agent"
)

func main() {
    config := newrelic.NewConfig("Test-Martini", "KEY_NAME")
    app, err := newrelic.NewApplication(config)
    if err != nil {
        log.Fatal(err.Error())
    }

    m := martini.Classic()
    m.Get("/build", func() string {
        return "12345"
    })
    m.Get("/hello/:p", func(params martini.Params) string {
        return "Hello " + params["p"]
    })
    m.Get("/slow", func() string {
        time.Sleep(3 * time.Second)
        return "OK"
    })
    m.Get("/custom", func(w http.ResponseWriter, req *http.Request) string {
        if txn, ok := w.(newrelic.Transaction); ok {
            txn.AddAttribute("CustomKey", "CustomValue")
            txn.SetName("CustomName")
            txn.NoticeError(errors.New("CustomError"))
        } else {
            // This triggers, but shouldn't.
            log.Println("Couldn't assert the type.")
        }

        return "custom things have happened"
    })

    http.ListenAndServe(":3000", newrelic.WrapHTTPHandler(app, m))
}
```

### DIFF
--- a/instrumentation.go
+++ b/instrumentation.go
@@ -38,6 +38,29 @@ func WrapHandleFunc(app Application, pattern string, handler func(http.ResponseW
 	return p, func(w http.ResponseWriter, r *http.Request) { h.ServeHTTP(w, r) }
 }
 
+// HTTPHandler allows us to satisfy the HTTPHandler interface
+type HTTPHandler struct {
+	app             Application
+	OriginalHandler http.Handler
+}
+
+// ServeHTTP satisfies the interface, kicks off a new Transaction using the requested
+// path. The transaction is passed as the ResponseWriter.
+func (handler *HTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	txn := handler.app.StartTransaction(r.URL.Path, w, r)
+	defer txn.End()
+
+	handler.OriginalHandler.ServeHTTP(txn, r)
+}
+
+// WrapHTTPHandler allows wrapping a ListenAndServe call.
+func WrapHTTPHandler(app Application, handler http.Handler) http.Handler {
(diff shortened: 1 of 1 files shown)


polds:
Hmm.. That indeed is a good catch. I wonder at the overhead of adding NewRelic as a middleware. Most http routers don't (easily) allow you to set the invocation order. So I feel that if you put NewRelic into a middleware you won't likely get the _whole_ picture of a request.

Someone correct me if I'm wrong.

sergeylanzman:
Yes it's true.
But if only a 0.001% difference in response time. That most of the time goes on http hendler

sergeylanzman:
Yes it's true.
But if only a 0.001% difference in response time. That most of the time goes on http hendler

willnewrelic:
Hi @polds 

Unfortunately, as @sergeylanzman mentioned, using the path as the transaction name can be dangerous in situations where path fields contain dynamic values.  As a result, we are going to decline this PR.  We appreciate your desire to contribute back nonetheless!

Our strategy is going to be to add framework specific packages to make instrumentation easier.

Sorry for the long delayed response.

PR #12, opened 2016-08-24
Override High Security in custom events

VadimBelov:
* Addition of flags for each of the features described [here](https://docs.newrelic.com/docs/agents/manage-apm-agents/configuration/high-security-mode#version2description)
* When turned on - each flag overrides the High Security mode in context of the given feature
* In this implementation Custom Events are sent even if High Security mode is enabled

### DIFF
--- a/config.go
+++ b/config.go
@@ -7,6 +7,17 @@ import (
 	"strings"
 )
 
+//
+type HighSecurityFeature uint16
+const (
+	EnabledHttpParam HighSecurityFeature = 1 << iota	// 1
+	EnabledMessageQueueParam				// 2
+	EnabledRawQueryStatement				// 4
+	EnabledUserAttribute					// 8
+	EnabledNoticeErrorAttribute				// 16
+	EnabledCustomEvents					// 32
+)
+
 // Config contains Application and Transaction behavior settings.
 // Use NewConfig to create a Config with proper defaults.
 type Config struct {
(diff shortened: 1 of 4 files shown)


victorNewRelic:
Hi Vadim,
I'm Victor, Product Manager for the Go Agent. Thank you for your PR. I've been discussing this internally with the Security and Engineering teams to work through what it means to disable the limitations around Custom Events when High Security Mode is enabled.

[High Security Mode (HSM)](https://docs.newrelic.com/docs/agents/manage-apm-agents/configuration/high-security-mode) is a setting that prevents possibly sensitive data from being sent to New Relic. This includes not sending data for HTTP parameters, custom parameters, raw queries, etc. since sensitive user information can be included in each of those categories.

Custom Events falls under the HSM restriction because of the potential for sensitive data to be sent to New Relic as part of the Custom Event payload. Changing the Go Agent to allow Custom Events in HSM would break the guarantees and contractual commitments New Relic has in place when a customer has opted to run under HSM.

We recognize that you have a valid need for sending Custom Events while under HSM. To safely provide the functionality without breaking guarantees, New Relic needs to take a broader approach by addressing how we handle this across all New Relic Agents.

I hope this explanation provides you the context you need for understanding why we're not accepting this PR. Since you have a support ticket in for this as well, I'll associate your ticket with the Feature Request we have for enabling Custom Events under HSM.

An option that you may wish to explore is filing a support ticket to have HSM disabled on your account so this restriction does not exist.

Best regards,
Victor Soares

PR #14, opened 2016-09-01
Correct StartTime struct field name in guide

sethhochberg:
This PR corrects the struct field name for `StartTime` in the example documentation for external segments. `Start` is not correct, the correct field name is `StartTime` as reflected in the type definition here: https://github.com/newrelic/go-agent/blob/master/segments.go#L45


willnewrelic:
Hi sethhochberg

Good catch!  We just caught this ourselves:  You will see it fixed in our next release.

I'm going to close this PR:  We appreciate your help.

PR #15, opened 2016-09-01
Don't Consume errors if Error Collector is disabled

VadimBelov:
Don't call to Consume (internal_txn.go:End()) on errors if Error Collector is disabled


willnewrelic:
Hi VadimBelov

Just released version 1.2.  This version has a fix:  Error metrics are no longer created when `ErrorCollector.Enabled` is set to false.  I am hopeful that this solves your issue!

Closing this PR:  Please reopen it if problems persist.

PR #18, opened 2016-10-19
Add missing word in description for external segments

abeltay:


willnewrelic:
Hi abeltay

Thanks for the fix!  This has been merged into our internal working master.  It will ship in our next release.

PR #24, opened 2016-11-11
Custom stack traces

fgrosse:
This pull request adds two new optional interfaces that errors may implement to customize how they are reported to newrelic. Additionally we added integration with [Travis-CI][4].There are no breaks in backwards compatibility for any exported types (considering the `newrelic/go-agent/internal` not to be exported).

## `stackTracer` interface

With this PR callers of `txn.NoticeError(err)` can now pass in an error type that can optionally implement the following interface to report custom stack trace information:

```go
// A stackTracer is an error type that can return information about the
// stacktrace at which it was created.
//
// The stackTracer interface is not exported by this package, but is considered
// to be part of the stable public API.
type stackTracer interface {
	StackTrace() []runtime.Frame
}
```

The problem this solves are error values which have been returned back up the stack before they are actually reported via `txn.NoticeError`. This happens quite frequently unless the transaction is passed as an extra argument throughout the stack which is seldomly acceptable.
Instead it is often much more practical to save the stack trace of the creation of the error so it can be reported at any later point in time. A library that implements this strategy already is [github.com/pkg/errors][2]. It is quite easy to wrap an error returned from `errors.Wrap(…)` in another error type that implements the `stackTracer` interface.

## `errorClasser` interface

Another unexported interface that is added with this PR is the `errorClasser` (following [the go naming conventions for interfaces][3]):

```go
// errorClasser is an error type that has a custom error class on New Relic.
//
// The errorClasser interface is not exported by this package, but is considered
// to be part of the stable public API.
type errorClasser interface {
	ErrorClass() string
}
```

This is mainly useful in order to wrap the error inside other error types without loosing the name of the original error type in New Relic.

## Example implementation

The following is just a copy of the `Example` for new integration for github.com/pkg/errors

```go
package nrerrors_test

import (
	"fmt"
	"path/filepath"

	"github.com/newrelic/go-agent/_integrations/nrerrors"
	"github.com/pkg/errors"
)

// Example illustrates how a StackTracer can be created from a wrapped error
// and how the according stack trace would look like. Note that since this
// example is compiled and executed automatically by the testing package the
// produced stack trace is a bit deeper than shown here in the output.
// To keep the example clean and reproducible only the first 3 stack frames
// are shown.
func Example() {
	err := doSomething()
	if err != nil {
		fmt.Println("ERROR:", err)
		err = nrerrors.WithStackTrace(err)
		// at this point we would usually call txn.Notice(err)
		st := err.(*nrerrors.StackTracer).StackTrace()
		for _, f := range st[:3] {
			fmt.Printf("%v:%v %v\n", filepath.Base(f.File), f.Line, f.Function)
		}
	}

	// Output:
	// ERROR: it failed: something went wrong
	// example_test.go:41 anotherFunction
	// example_test.go:37 doSomething
	// example_test.go:18 Example
}

func doSomething() error {
	return errors.Wrap(anotherFunction(), "it failed")
}

func anotherFunction() error {
	return errors.New("something went wrong")
}

```

## Compatibility notice

Please note that the interface depends on the `runtime.Frame` type of the standard library which is only available in go 1.7 and above. We use build tags to still be able to build the go-agent on older versions. In order to verify that the go-agent is still working correctly on go 1.3 and above a .travis.yml was added to run tests on multiple versions using [Travis CI][1]. You will have to activate the integration in the newrelic repository yourself once the PR is merged.

[1]: https://travis-ci.org/classmarkets/go-agent
[2]: https://github.com/pkg/errors
[3]: https://golang.org/doc/effective_go.html#interface-names
[4]: https://travis-ci.org


willnewrelic:
Hi fgrosse

Thanks for this PR.  These changes look compelling:   I'm going to leave this open and add the "track-interest" label. I encourage all who would like this change to add a comment or thumbs up.

fgrosse:
I just saw there are now conflicts due to a change in the JSON structure of the stack frames. 
I am going to merge the upstream master into my branch to resolve the conflicts so my branch now also  uses the new format from https://source.datanerd.us/agents/agent-specs/blob/master/Stack-Traces.md

fgrosse:
It seems like https://source.datanerd.us/agents/agent-specs/blob/master/Stack-Traces.md isn't public. Is this misconfiguration or intended? If this is a private repo you may want to remove comments referencing this repo. For now I will just mirror the [current implementation of `StackTrace.WriteJSON`][1].

[1]: https://github.com/newrelic/go-agent/blob/7d12ae2201fc160e486197614a6f65afcf3f8170/internal/stacktrace.go#L51

fxposter:
Ability to support passing stacktrace would be really cool, because usually you want to do `NoticeError` not at the point where you first get the error, but somewhere at the higher level, far from the place where the original error was returned.

fgrosse:
Hey @willnewrelic , Having the integration directly in this repository sounds like a great idea so I just converted the example implementation from my initial pull request description into a full integration as requested. The `example_test.go` can also be run via go test and verifies that the produced output matches the documented expectation. However keep in mind that those tests are not included in `go test github.com/newrelic/go-agent/...`. I will update the travis.yml accordingly.

I will take some more time to think about your question regarding the `stackTracer` interface and answer to that one separately later.

P.S.: I updated the PR description by replacing the old executable example section with a copy of the new `example.go` which imrpoves readability of this PR quite a bit :smile:

fgrosse:
### About the `stacktracer` interface

I gave this some thought when I wrote the `stackTracer` interface and I decided against using `[]uintptr` for the following reasons:

- the interface should be **useful in all contexts** (not only in the newrelic one)
  As the caller of `StackTrace` I most likely want to know about the stack frames immediately. Returning pointers that need to be resolved manually is tedious and forces me to import the `runtime` package.
- `runtime.FuncForPC` is not **straight forward**. The newrelic package itself still has [a `TODO`][3] to understand how exactly this function is supposed to be used.
- **stack frames are immutable**. That means that once I resolved a `pc` to a frame its value (file, line, function) cannot change. There is no point in resolving the same stack frame multiple times
- there is a **type within the standard library** that represents stack frames (namely [`runtime.Frame`][4]) so we can use this to build generic interfaces that do not depend on custom types. Furthermore by accepting this interface others are encouraged to implement and use it as well.

I couldn't really come up with good arguments to use `uintptr`. If this was about using `uintptr` vs. a custom `newrelic.StackFrame` I would absolutely agree that `uintptr` is much better since it allows to create a universally useful interface that doesn't depend on `newrelic`. However having the standard library type gives us the best of both worlds so the only counter argument that I can think of is the minimum required version.

### About depending on go 1.7+

The go community is in the fortunate position that upgrading from one version to another (like from 1.3 to 1.7) is very easy due to the compatibility promises of the standard library and the easy setup. Therefore I actually doubt that there are very many people who are activly developing a product with go but are stuck to an old version. If such use cases does exist than it is important to keep in mind that  this PR would not break their code - they could just not use the new functionality.

### About other error packages

I was looking for other error packages and what interface they use to return a stack trace.
So far I only found a single package: [github.com/go-errors/errors][1] which also [returns the stack trace][2] as a slice of `StackFrame` structs. If this package would actually use [the new type from the standard library][4] than this function could easily be used by others without having to import that specific package.

[1]: https://github.com/go-errors/errors
[2]: https://godoc.org/github.com/go-errors/errors#Error.StackFrames
[3]: https://github.com/newrelic/go-agent/blob/7d12ae2201fc160e486197614a6f65afcf3f8170/internal/stacktrace.go#L36
[4]: https://godoc.org/runtime#Frame

pschultz:
It might be worth considering that other error tracking libraries can benefit from the interface suggested here as well.

I glanced over the godocs for the clients for [Airbrake](https://godoc.org/gopkg.in/airbrake/gobrake.v2), [Rollbar](https://godoc.org/github.com/stvp/rollbar), [Raygun](https://godoc.org/github.com/MindscapeHQ/raygun4go), and [Sentry](https://godoc.org/github.com/getsentry/raven-go) (the package is called raven).

`raygun4go` seems to only handle panics, so it doesn't really compare to the others. `rollbar` and `raven` allow callers to pass their respective custom trace types, which are a superset of `runtime.Frame`. `gobreak` accepts a depth parameter and presumably computes the stack automatically.

Whether or not they are willing to support it is obviously another question, but at a glance they could consume either interface.

I suppose the question is who should be burdened with converting `[]uintptr` to `[]runtime.Frame`, the implementor or the consumer of the interface.

Also, the `[]runtime.Frame` interface can be mocked, and errors that carry `runtime.Frame`s can be serialized in a meaningful way. Neither works with `uintptr`s.

I personally vote for `[]runtime.Frame`.

willnewrelic:
@fgrosse @pschultz

Thanks for the detailed feedback.  We really appreciate it.

After much discussion and debate, we are inclined to use `[]uintptr` for the
stack traces for the following reasons:

* We seek to minimize the agent's cost of generating a stack trace.  Not all generated stack traces are sent to our servers, so there is benefit in deferring as much work until JSON creation time as possible.  Although we could have multiple internal stack trace types (one for errors, and another one for traces), we do not want this extra complexity.

* If we rely on `runtime.Frame`, most consumers would likely only populate the currently used `Function`, `File`, and `Line` fields, which would limit our ability to use other fields such as `PC` in the future.

* Simpler integration with integration with pkg/errors (since no manual `runtime.Frame` creation).

Thus, we are inclined to use:

```
type ErrorClasser interface {
	ErrorClass() string
}

type StackTracer interface {
	StackTrace() []uintptr
}
```

We are tempted to make these interfaces exported so that they are more easily
discoverable though godocs.

willnewrelic:
Hi @fgrosse @pschultz

I have created https://github.com/newrelic/go-agent/pull/34  with our latest thinking on this functionality.  We invite feedback.

Once again, we really appreciate your help and interest!

willnewrelic:
Support for ErrorClasser, StackTracer, and pkg/errors released as part of 1.6.0!  Thanks again for the help and feedback!

PR #25, opened 2016-11-14
Update memtotal_darwin.go

urunimi:
Add a comma

### DIFF
--- a/internal/sysinfo/memtotal_darwin.go
+++ b/internal/sysinfo/memtotal_darwin.go
@@ -7,7 +7,7 @@ import (
 
 // PhysicalMemoryBytes returns the total amount of host memory.
 func PhysicalMemoryBytes() (uint64, error) {
-	mib := []int32{6 /* CTL_HW */, 24 /* HW_MEMSIZE */}
+	mib := []int32{6 /* CTL_HW */, 24 /* HW_MEMSIZE */,}
 
 	buf := make([]byte, 8)
 	bufLen := uintptr(8)


urunimi:
It is not possible to build without the comma.

internal/sysinfo/memtotal_darwin.go
```diff
@@ -7,7 +7,7 @@ import (
 
 // PhysicalMemoryBytes returns the total amount of host memory.
 func PhysicalMemoryBytes() (uint64, error) {
-	mib := []int32{6 /* CTL_HW */, 24 /* HW_MEMSIZE */}
+	mib := []int32{6 /* CTL_HW */, 24 /* HW_MEMSIZE */,}
```
fgrosse: Why wouldn't this build without the comma, which go version do you use?

urunimi: I am using 1.7.3

fgrosse: The comma would only [be required][1] if there was a newline before the closing bracket.
What compilation error do you get when building this without the comma?

This is my output:
```
$ go version
go version go1.7.3 linux/amd64

$ GOOS=darwin GOARCH=arm64 go build -v ./internal/sysinfo/memtotal_darwin.go
runtime/internal/sys
runtime/internal/atomic
runtime
internal/race
errors
sync/atomic
sync
syscall
command-line-arguments

$ echo $?                                                                   
0
```

[1]: https://dave.cheney.net/2014/10/04/that-trailing-comma

urunimi: Oh, it seems that formatter added new lines. Thanks!

fgrosse: no problem :D

PR #28, opened 2016-12-08
Gin Framework Support

willnewrelic:
WARNING: THIS CODE IS IN DEVELOPMENT. IT IS NOT A FINAL NEW RELIC PRODUCT OR OFFERING. DO NOT USE IN PRODUCTION. USE AT YOUR OWN RISK. TERMS AND CONDITIONS APPLY - SEE LICENSE FILE.

Hi All

This is pre-release support for the Gin framework. Let us know what you think. We are eager to get your feedback!

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -5,46 +5,235 @@ under the following BSD license:
 
 -------------------------------------------------------------------------------
 
-All components of this product are Copyright (c) 2016 New Relic, Inc.  All
-rights reserved.
+New Relic Pre-Release Program Terms
 
-Certain inventions disclosed in this file may be claimed within patents owned or
-patent applications filed by New Relic, Inc. or third parties.
+These terms apply to the associated software. By downloading or installing the
+software, you signify that you have read, understood, and accept these terms
+(the “Agreement”) governing your participation in New Relic’s program for making
+beta and pre-release versions of its software, services, and features available
+for testing and evaluation purposes (the “Pre-Release Program”). This Agreement
+is entered into by and between New Relic, Inc. (“New Relic”) and you on behalf
+of yourself or the entity associated with your New Relic account (collectively
+referred to herein as “you” or “your,”) as of the date you download the
+applicable software (the “Effective Date”). You represent and warrant that you
+have the legal authority to enter into this Agreement personally or on behalf of
+the company or organization associated with your New Relic account and such
+Agreement shall be valid and binding on such company or organization (“Eligible
+Organizations”).
 
-Subject to the terms of this notice, New Relic grants you a nonexclusive,
-nontransferable license, without the right to sublicense, to (a) install and
-execute one copy of these files on any number of workstations owned or
-controlled by you and (b) distribute verbatim copies of these files to third
-parties.  You may install, execute, and distribute these files and their
-contents only in conjunction with your direct use of New Relic’s services.
-These files and their contents shall not be used in conjunction with any other
-product or software, including but not limited to those that may compete with
-any New Relic product, feature, or software.  As a condition to the foregoing
-grant, you must provide this notice along with each copy you distribute and you
-must not remove, alter, or obscure this notice.  In the event you submit or
-provide any feedback, code, pull requests, or suggestions to New Relic you
-hereby grant New Relic a worldwide, non-exclusive, irrevocable, transferrable,
-fully paid-up license to use the code, algorithms, patents, and ideas therein in
-our products.
+IF YOU DO NOT AGREE TO THESE TERMS OR CANNOT AGREE TO THESE TERMS ON BEHALF OF
+AN ELIGIBLE ORGANIZATION, DO NOT USE THE SOFTWARE AND/OR BETA SERVICE.
 
-All other use, reproduction, modification, distribution, or other exploitation
-of these files is strictly prohibited, except as may be set forth in a separate
-written license agreement between you and New Relic.  The terms of any such
-license agreement will control over this notice.  The license stated above will
-be automatically terminated and revoked if you exceed its scope or violate any
-of the terms of this notice.
+1. Introduction.
 
-This License does not grant permission to use the trade names, trademarks,
-service marks, or product names of New Relic, except as required for reasonable
-and customary use in describing the origin of this file and reproducing the
-content of this notice.  You may not mark or brand this file with any trade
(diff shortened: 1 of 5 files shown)


willnewrelic:
Is the `v1` subdirectory the best way to handle framework versioning?

willnewrelic:
Should the Transaction be added to the gin Context so that it can be accessed in the handlers?

jasonab:
I would love to see some error integration into this middleware, so that if I do AbortWithError, that is automatically picked up by the Gin integration and flashed to NR.

rodriguezgustavo:
Hi @willnewrelic . We using your middleware. It works great. As you mention in your comment, it would be very useful to have the transaction available in the handlers. Is it possible to add it to the gin.Context? Without the transaction we cannot show segments and other useful stuff. On the other hand, do you now when it would be released? Thanks

willnewrelic:
@rodriguezgustavo  Just added a new commit that adds a function `nrgin.Transaction` to access the transaction.  Enjoy!  Example use:
```go
func handlerAccessTransaction(c *gin.Context) {
	if txn := nrgin.Transaction(c); nil != txn {
		txn.SetName("custom-name")
	}
	c.Writer.WriteString("changed the name of the transaction!")
}
```
@jasonab Interesting idea.  Upon investigating, I was surprised to see that using `AbortWithError` does not automatically record errors when the code is greater than 400 (and not 404).  It should be intercepted...  Looks like the `gin.Context.Status` function uses `Context.writermem` and not `Context.Writer`, which I don't understand.  I will make a PR to gin to see what they think.

willnewrelic:
Looks like someone already made a (broken) PR to gin to fix `Context.Status`:  https://github.com/gin-gonic/gin/pull/625

ingshtrom:
What is the status of this? How does this get in and then also how does it get "approved" for production use?

rodriguezgustavo:
HI, we were using this code in production for more than 1 month without any issue (APIs with more than 1 million RPM). Do you know when will you merge this PR to master?

LeoAdamek:
I'm also very interested in when this integration will become generally available. It sounds like its pretty stable according to @rodriguezgustavo , I'd be willing to try this pre-release out on a pre-release app but I'm unable to get the integration to pull into my vendor dir so can't try it out.

ingshtrom:
@rodriguezgustavo could you elaborate on how you are running this? How did you set this up with a basic setup of `gin`? tia

LeoAdamek:
I don't know how @rodriguezgustavo has done it, but what I've done so far myself is implement a simple Middleware for Gin which is added to the handler chain if a `NEWRELIC_KEY` environment variable is set.

This middleware for each request creates the NR transaction, gives it a default name and properties and then puts the transaction on the gin context so the request handler can annotate it further.

rodriguezgustavo:
Hi @LeoAdamek,

Yo can use nrgin.Middleware(newrelicApp) to avoid writting your custom middleware. Then in your gingonic handlers you can access to the transaccion calling to nrgin.Transaction(*gin.Context)

willnewrelic:
Hi All

This work has been merged into master and is now released.

sudo-suhas:
@willnewrelic Not sure if I should open a new issue but status codes are not captured correctly in newrelic.

For this sample code:

```go
ctx.JSON(
	http.StatusUnprocessableEntity, // 422
	util.ErrorResp(validationErrMsg, vErrs),
)
```

... newrelic shows the `httpResponseCode` as 200:

image

And a header which was sent in the request is missing too.

AndyEsser:
Does New Relic correctly capture the HTTP Status Code now of requests? I have data populating in my dashboard, but when a response is 400 or 403 I don't see any errors displayed

crenshaw:
Thanks @AndyEsser for your question.  I will submit a ticket for this question and get back to you.

periclesroo:
Are there any updates on what @sudo-suhas reported? I have pretty much the same issue, we raise a 500, the response code is 500 but new relic reports a 200.

sudo-suhas:
So here's the work around I used:
- Turn off tracking for the response code using the config passed to `newrelic.NewApplication`.
- Add a middleware to track the response code manually.
```go
cfg := newrelic.NewConfig("myapp-"+config.App.Env, config.App.NewrelicLicense)

// With gin, newrelic is not able to capture HTTP response code correctly.
// So we disable the defaults as shown here -
// https://github.com/newrelic/go-agent/blob/master/GUIDE.md#attributes
// and track it ourself.
cfg.Attributes.Exclude = append(cfg.Attributes.Exclude, newrelic.AttributeResponseCode)

var err error
if Agent, err = newrelic.NewApplication(cfg); err != nil {
    log.WithError(err).Panic("Newrelic agent spawn failed!")
}

// trackInNR tracks additional attributes for all requests including errors, if
// any. It only tracks the HTTP response status code as newrelic seems unable to
// track it correctly.
func trackInNR(ctx *gin.Context) {
	ctx.Next()

	if txn := nrgin.Transaction(ctx); txn != nil {
		err := txn.AddAttribute(newrelic.AttributeResponseCode, ctx.Writer.Status())
		if err != nil {
			log.WithError(err).Warn("Got error while trying to track attribute")
		}

		errMsgs := ctx.Errors.ByType(gin.ErrorTypePublic | gin.ErrorTypePrivate)
		if len(errMsgs) > 0 {
			qryStr := ctx.Request.URL.RawQuery
			txn.AddAttribute("queryParams", qryStr)

			for _, errMsg := range errMsgs {
				if err := txn.NoticeError(errMsg.Err); err != nil {
					log.WithError(err).Warn("Got error while trying to track attribute")
				}
				// For now, not trying to send the meta to newrelic.
				// `txn.AddAttribute` will only accept and send string, numbers
				// and booleans. We'd need to flatten all the values recursively
				// which might be costly.
			}
		}
	}
}

// and use middleware
router.Use(nrgin.Middleware(monitor.Agent), trackInNR)
```

willnewrelic:
Hi all

Regarding the http response codes:  Yes, unfortunately using any `gin.Context` methods that delegate to `Context.Render` (such as `context.JSON`) will result in missing response codes.  This is a consequence of the `Context.Status` method not using the `Writer` field which our instrumentation has replaced. 

I have created a PR against Gin to get this fixed here:  https://github.com/gin-gonic/gin/pull/1606

Hoping that gets some love soon!

PR #34, opened 2017-01-12
Support for ErrorClasser, StackTracer, and pkg/errors

willnewrelic:
WARNING: THIS CODE IS IN DEVELOPMENT. IT IS NOT A FINAL NEW RELIC PRODUCT OR OFFERING. DO NOT USE IN PRODUCTION. USE AT YOUR OWN RISK. TERMS AND CONDITIONS APPLY - SEE LICENSE FILE.

Hi All

This is our current thinking about addressing the functionality requested in:
https://github.com/newrelic/go-agent/pull/24

 We are eager to get your feedback.  Some questions:

* Are there any downsides to making the interfaces `StackTracer` and `ErrorClasser` exported?
* Is `nrpkgerrors.Transform` a good name for the function designed to wrap `pkg/errors` errors?


willnewrelic:
Some questions:

* Are there any downsides to making the interfaces `StackTracer` and `ErrorClasser` exported?

* Is `nrpkgerrors.Transform` a good name for the function designed to wrap `pkg/errors` errors?

fxposter:
@willnewrelic I think that you should not require for people to use `nrpkgerrors.Transform(e)` in their code and do the transformation internally. Otherwise you require either:
1. Integrating go-agent deeply into application, so that every part of the app knows that it actually uses newrelic and need to depend on it.
2. Making each application to have a wrapper around go-agent which does transform and other application-related things related to errors.

IE: from the application perspective having a dependency on "pkg/errors" is fine, cause it is actually quite a generic and small thing, while making whole app do a newrelic-specific `Transform` (or requiring to build a wrapper to hide it in each app) does not sound nice for me.

pschultz:
I think exporting the interfaces is a good idea. There is a lot of precendent in the standard library such as json.Marshaler. As Will mentioned, it makes it discoverable in godocs, and it is a hint for future developers of the agent package that changing the interface affects not only the package itself but also its users.

Have you considered `nrpkgerrors.Wrap`?

fxposter:
That is the thing - you usually want your application to depend on pkg/errors freely, but you usually would not want to depend on nrpkgerrors directly. And if you do `nrpkgerrors.Transform()` just before `NoticeError` - what is the point of doing it manually every time, instead of library detecting that the error was wrapped?

pschultz:
If you are calling NoticeError at the point where the error occurs, you don't have to call nrpkgerrors.Transform because the automatic stacktrace is already correct.

nrpkgerrors.Transform is only necessary if NoticeError is called in some other place; then the automatic trace is irrelevant. Do you expect there to be many other places? In our applications NoticeError is called exactly once, in an HTTP middleware. Nothing else has to depend on nrpkgerrors. From this perspective it is acceptable as is.

newrelic could detect errors from github.com/pkg/errors, but I am not convinced that is okay to assume that this is the trace the user wants. For instance, if my SQL library returns an error wrapped by github.com/pkg/errors, I may not be interested in the trace within that library. Even worse, if that is a deep stack, I may loose all the calls in my application. nrpkgerrors.Transform makes that explicit, which I prefer.

willnewrelic:
Thanks for the comments.

`nrpkgerrors.Wrap` does seem better.  Will use!

We want the error transformation to be explicit:  We seek to give control to the consumer, and avoid an agent dependency on pkg/errors.  Therefore, we are committed to keeping nrpkgerrors as a separate package.

willnewrelic:
This has been released as part of 1.6.0!

PR #36, opened 2017-03-02
reduce memory consumption in collectorRequestInternal()

pseidemann:
hi,
I'm currently debugging a memory leak in my go app which uses the newrelic go agent. I saw that the agent uses a lot of memory so I looked into improving the agent's memory usage.
I hope this is useful to you.

- stop calling Bytes() in compress(), instead returning pointer to buffer, allowing direct usage for NewRequest()
- stop calling ReadAll(), instead passing Body to parseRequest() and use Decoder which allows streamed parsing without copying
- delete dead code compressEncode() and uncompressDecode()


willnewrelic:
Hi @pseidemann 

This looks great!  Thank you very much for the contribution!

Expect to see this change in our next release.

willnewrelic:
Hi @pseidemann 

Looks like there could be draining issues with this `json.NewDecoder(body).Decode(&r)` change:  See "json.Decoder does not drain HTTP connections properly" section on https://ahmetalpbalkan.com/blog/golang-json-decoder-pitfalls/

As a result, we intend to focus on the prevent-request-copy part of this PR for the next release.

rvanderwal-newrelic:
Thanks for your ideas, @pseidemann ! We incorporated most of these ideas in today's release.

PR #40, opened 2017-04-20
Add logger integration for inconshreveable/log15

jescochu:
This is an integration for a new logger we are trying out at https://github.com/inconshreveable/log15.

It is fairly straightforward except that unfortunately the lib exposes no easy way to get the current log level bit (and debug is the default log level). To cope with this, there are two helper methods that offer varying levels of control of the actual logger. The first would require a bool that tells whether or not we are in "debug mode" to satisfy the newrelic.Logger interface. StandardLogger(), like the version in the nrlogrus integration, will just instantiate a default logger.

https://github.com/inconshreveable/log15/issues/111

### DIFF
--- a/_integrations/nrlog15/nrlog15.go
+++ b/_integrations/nrlog15/nrlog15.go
@@ -45,15 +45,11 @@ func New(logger log.Logger, debug bool) newrelic.Logger {
 	}
 }
 
-func StandardLogger(debug bool) newrelic.Logger {
+func StandardLogger() newrelic.Logger {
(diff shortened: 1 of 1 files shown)


robertnewrelic:
@jescochu thanks for taking time to work on this and contribute to the project!  I like what I see and I think it looks good, however I sat down with the team and we've decided to not accept new contributions to our integrations at this point in time.

Accepting new integrations causes us to make sure compatibility with those projects do not break in the future.  If that decision changes we'll be sure to reopen this PR!

PR #49, opened 2017-06-09
change license length validation to check for minimum size

Caleb-An:
Recently we have had to upgrade our account license key, and all our go services stopped reporting. I've discovered that it is because our new key (which NewRelic provided) is 53 characters long, and there is an existing check for keys being exactly 40 characters. 

I've made this PR check for key lengths being above or equal to 40, because it seems unlikely that all new keys are 53 characters long; I'm guessing your new key generation generates keys at least 40 long.

### DIFF
--- a/config.go
+++ b/config.go
@@ -219,13 +219,13 @@ func NewConfig(appname, license string) Config {
 }
 
 const (
-	licenseLength = 40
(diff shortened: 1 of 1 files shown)


Caleb-An:
Turns out we received a malformed key, closing.

Issue #19, opened 2016-10-20
Support alternate method of having a "web" transaction


Ewan-GrandSlam:
In using this new-relic agent, I am only able to use an http.X type to classify a request as an http or web request - however, due not using the standard net/http library I am only able to track my transactions as "background" requests - which is alright however does not accurately reflect what the data is actually doing.

As an example if I use a framework such as fasthttp where it is essentially io.writer and io.reader  - I am unable to create a web transaction with the current agent - although I may have all the data needed to send a web transaction.

willnewrelic:
Hi Ewan-GrandSlam

Thanks for the idea.  I'm going to leave this open and add the "track-interest" label.  I encourage all who would like this change to add a comment or thumbs up.

kmhosny:
Hi there,

is there any update on how to track web transactions when using the non standard net/http such as fasthttp? what kind of information does the go-agent extract from http.Response and http.Request may be i can populate it by hand?

thanks

willnewrelic:
Hi

Thanks for all the interest.  I will investigate fasthttp.

You can turn a background transaction into a web transaction by providing an empty `http.Request` as a parameter to `StartTransaction`:

```go
txn := app.StartTransaction("notBackground", nil, &http.Request{})
```

greenboxal:
I have interest in this as well. I'll start a PR changing the way the library treats http requests to use generic interfaces that can be implemented by anyone. So in case of anyone using fasthttp (my case), it will be just a matter of implementing the correct interface on top of it.

greenboxal:
@kmhosny In case you need it, I just pushed https://github.com/greenboxal/nr-fasthttp. It depends on #56.

robertnewrelic:
Thanks to everyone for highlighting this issue!  @Ewan-GrandSlam @kmhosny @greenboxal 

This resulted in us opening a feature request to further look into the problem.  We'll also take your PR into consideration @greenboxal !

I'm going to close this issue since we'll be tracking it internally.  At this time I do not have an ETA when to expect this fix.

willnewrelic:
We are going to do this!

Your feedback wanted:

https://github.com/newrelic/go-agent/issues/19

PR #56, opened 2017-07-22
Add generic HTTP interface

greenboxal:
This enables HTTP libraries that aren't based on net/http to be used.

Unfortunately, it breaks ExternalSegment interface. Do anyone have a better idea that doesn't do this?

Addresses #19

### DIFF
--- a/application.go
+++ b/application.go
@@ -15,7 +15,7 @@ type Application interface {
 	// * If an http.ResponseWriter is provided then the Transaction can be
 	//   used in its place.  This allows instrumentation of the response
 	//   code and response headers.
-	StartTransaction(name string, w http.ResponseWriter, r *http.Request) Transaction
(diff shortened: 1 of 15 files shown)


robertnewrelic:
@greenboxal Thanks for this PR!  We'll take this into consideration when we begin looking at the feature request filed relating to issue #19 .

willnewrelic:
Hi @greenboxal 

We are planning on adding a `Transaction.SetWebRequest` method in the next release to address this feature request.

How does this look:  https://github.com/newrelic/go-agent/issues/78

willnewrelic:
This functionality has now been released in 2.2.0!

Please check out the new `Transaction.SetWebRequest` method!

https://godoc.org/github.com/newrelic/go-agent#WebRequest

PR #63, opened 2017-12-12
Get AppID for an application

aarthi184:
Retrieve the NewRelic AppID from the go-agent instead of hitting the REST API to get the Application ID.

**Problem:**
Need to tag deployments from the service itself.
A HTTP call needs to be made to get the AppID from the REST API which will list all the applications and filter by name.

**Solution:**
Since the go-agent maintains the AppID after connection has been established to the NewRelic servers, retrieving it from here would reduce JSON encode/decode and a HTTP call.

### DIFF
--- a/application.go
+++ b/application.go
@@ -44,6 +44,11 @@ type Application interface {
 	// application is connected successfully.
 	WaitForConnection(timeout time.Duration) error
 
+	// ID returns the application's ID in New Relic. It will return the ID of the
+	// application only after connection has been established.
+	// Returns an error, if connection has not been established yet.
+	ID() (string, error)
+
 	// Shutdown flushes data to New Relic's servers and stops all
 	// agent-related goroutines managing this application.  After Shutdown
 	// is called, the application is disabled and no more data will be
--- a/internal_app.go
+++ b/internal_app.go
@@ -374,6 +374,24 @@ func (app *app) WaitForConnection(timeout time.Duration) error {
 	}
 }
 
+func (app *app) ID() (string, error) {
+	if !app.config.Enabled {
+		return "", fmt.Errorf("config not enabled")
+	}
+
+	run, err := app.getState()
+	if nil != err {
+		return "", fmt.Errorf("not connected. %s", err)
+	}
+
+	if run.RunID != "" {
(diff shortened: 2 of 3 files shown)


worldofprasanna:
We need to validate our connection from go-agent to new relic is successful or not. This is needed for the health endpoint which we build. This pull request which fetches the AppID would be good for us to know that go-agent can successfully connect to new relic server.

Can someone please verify the pull request and merge it. Thanks.

robertnewrelic:
Thank you for this PR @aarthi184 !  We will take this change into consideration.  The team is extra cautious of PRs that make changes to the public API, but know this is on our radar and will be looked into.

@worldofprasanna You mentioned needing to validate your connection with New Relic was successful, but not explicitly needing the Application ID.  Would waiting for a connection to be established before proceeding be helpful to you?  There is an example in the repo of doing so: https://github.com/newrelic/go-agent/blob/master/examples/short-lived-process/main.go#L27.  Please let me know if I have interpreted your need incorrectly.

worldofprasanna:
@robertnewrelic Thanks Robert. This one works for me 👍

aarthi184:
@robertnewrelic @crenshaw Can you let me know why this PR was closed? Though the conversation digressed to health checks, the original problem still exists.

astormnewrelic:
@aarthi184 Oh no, our apologies!  We didn't mean to lose the original PR in the shuffle.  :( 

We appreciate the time/effort you put into making this pull request, we always love to see 1. What people are doing with our agent and 2. How folks are taking advantage of a source available agent to solve their own problems. 

That said -- anytime we add a method call to a public interface it becomes something we need to support ~forever~ for a significant amount of time for all our users.   Adding this public `ID` method to the application interface is something we can't do at this time.  

We have created an internal issue to track the idea (both of adding the ID, as well as tagging deployments for the service itself without a REST API call), and while this PR is closed it's still available and searchable for folks who might want to incorporate it into their own monitored Go Applications. 

Thank you for understanding, and if this is causing you serious issues please don't hesitate to get in touch via the usual customer support channels ([support.newrelic.com](http://support.newrelic.com), or email (support@newrelic.com))

PR #64, opened 2018-01-23
Fix debug level in logrus integration

paddycarey:
This PR updates the logrus log level when debug is used.


willnewrelic:
Hi @paddycarey 

Thanks for bringing this to our attention!  This will be fixed in our next release!

PR #65, opened 2018-02-04
Change specific gin context struct parameter by go context interface

rodriguezgustavo:
Since gin context implements go context and we don't need gin context specific info inside this function it is better to use the interface instead the gin context. This change doesn't break current usages of this function and will provide flexibility in clients

### DIFF
--- a/_integrations/nrgin/v1/nrgin.go
+++ b/_integrations/nrgin/v1/nrgin.go
@@ -8,6 +8,7 @@ package nrgin
 import (
 	"net/http"
 
+	"context"
 	"github.com/gin-gonic/gin"
 	"github.com/newrelic/go-agent"
 	"github.com/newrelic/go-agent/internal"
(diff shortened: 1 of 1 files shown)


willnewrelic:
Hi @rodriguezgustavo 

Sorry for the long delay:  The Go agent is going to get more attention going forward!

We plan on fixing this in the next release.  We're thinking of using the code below, as it would for all Go versions.  This look okay?

```go
var (
	ctxKey = "newRelicTransaction"
)

// Context avoids making this package 1.7+ specific.
type Context interface {
	Value(key interface{}) interface{}
}

// Transaction returns the transaction stored inside the context, or nil if not
// found.
func Transaction(c Context) newrelic.Transaction {
	if v := c.Value(ctxKey); nil != v {
		if txn, ok := v.(newrelic.Transaction); ok {
			return txn
		}
	}
	return nil
}
```

Question for you:  `newrelic.NewContext` and `newrelic.FromContext` do not play well with the `nrgin.Transaction`:  The `newrelic` package functions use a custom key type (`type contextKeyType struct{}` as is recommended)  and `nrgin` uses a string key because `gin.Context.Set` requires a string.

This lack of compatibility is unfortunate because we plan on introducing more instrumentation which uses `newrelic.FromContext` [example here](https://github.com/newrelic/go-agent/issues/79):  This instrumentation wouldn't work for people using `nrgin.Middleware`.  Should we change use a string context key everywhere?  I suppose `var contextKey = "newRelicContextKeyPleaseDontCollide"` is unlikely to collide with anything!

willnewrelic:
Hi @rodriguezgustavo

We have just released 2.2.0 with this fixed!

We encourage you to check out https://godoc.org/github.com/newrelic/go-agent/_integrations/nrgin/v1#Transaction

PR #66, opened 2018-03-02
gorilla: carry the transaction within the request context

caarlos0:
So the user is able to easily get the current transaction in any method that has the context.

It should be useful to add the external calls as well sql calls :)

### DIFF
--- a/_integrations/nrgorilla/v1/nrgorilla.go
+++ b/_integrations/nrgorilla/v1/nrgorilla.go
@@ -3,6 +3,7 @@
 package nrgorilla
 
 import (
+	"context"
 	"net/http"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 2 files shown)


gaillard:
I am doing something similar in our application, however it would be great if the newrelic team, as part of the agent, could handle starting a new tx or something if it gets concurrent segments. 
That would make passing in the context that ends up in multiple goroutine to give reliable numbers, but I haven't found a way to do it yet.

cjcjameson:
@willnewrelic has a branch on this repo with something about gorilla...?

crenshaw:
@caarlos0 -- This note is overdue.  When you first put in this PR, my team's interest was immediately piqued by such a useful feature.  That said, it's been five months.  Hopefully I can make up for this delay by thanking you for this contribution's appearance in the official 2.1 release.  See: https://github.com/newrelic/go-agent/blob/master/CHANGELOG.md#210

caarlos0:
🤷‍♂️

PR #71, opened 2018-05-23
wrap the root

malmuzaini:
Implement a wrapper for serveHTTP so all of handleFunc will be wrapped, instead of wrapping each one.

### DIFF
--- a/instrumentation.go
+++ b/instrumentation.go
@@ -31,11 +31,25 @@ func WrapHandle(app Application, pattern string, handler http.Handler) (string, 
 	})
 }
 
+func WrapAllHandle(app Application, handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		txn := app.StartTransaction(r.URL.Path, w, r)
+		defer txn.End()
+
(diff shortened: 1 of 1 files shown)


malmuzaini:
the name of the functions can be changed, but the main idea is to provide a wrapper for the "/" root, so that when the errors are logged to NewRelic it won't be logged with root, but with. actual request url.

instrumentation.go
```diff
@@ -31,13 +31,27 @@ func WrapHandle(app Application, pattern string, handler http.Handler) (string,
 	})
 }
 
+func WrapAllHandle(app Application, handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		txn := app.StartTransaction(r.URL.Path, w, r)
```
gaillard: I doubt you want all the dynamic parts of the path to be in a transaction name?

willnewrelic:
Hi @malmuzaini 

We appreciate your desire to contribute!  Unfortunately, as @gaillard indicates, our public instrumentation does not use the url path since the number of transaction names could explode if the path contains dynamic fields.

PR #74, opened 2018-07-10
idea for how to check for an invalid license key

cjcjameson:
# Proposal for new behavior: Invalid License Key

## Current Behavior

If the user configures the Application with a 40-digit license key, there's no
further validation at Application-creation time.
Then if they run the application, the behavior `StartTransaction()` is
indistinguishable from if they had set `enabled: false`. 

I suggest that a user of this library should have a way to distinguish between
a disabled app versus a misconfigured one.

## Proposal 1

A new public method on internal_app.app:

```
func (app *app) ValidateConfig() error
```

The implementation of this would check that the license key corresponds to an
actual NewRelic account, perhaps by hitting some 'ping' endpoint you might have
available, or by starting a demo transaction.

We wouldn't want to add it to the Application interface in a minor version
bump, because the interface is used as a return value from `NewApplication`
(rather than a concrete struct).
It could affect clients or their tests that use it.

## Proposal 2

Could this be checked as part of `newAppRun()`?

### PR vs. Issues

I'm filing this as a PR, but I'd rather have filed an issue. Is there a reason
issues aren't enabled for your repo?

### DIFF
--- a/invalid_license_key_proposal.md
+++ b/invalid_license_key_proposal.md
@@ -1,0 +1,37 @@
+# Proposal for new behavior: Invalid License Key
+
+## Current Behavior
+
+If the user configures the Application with a 40-digit license key, there's no
+further validation at Application-creation time.
+Then if they run the application, the behavior `StartTransaction()` is
+indistinguishable from if they had set `enabled: false`. 
+
+I suggest that a user of this library should have a way to distinguish between
+a disabled app versus a misconfigured one.
+
+## Proposal 1
+
+A new public method on internal_app.app:
+
+```
+func (app *app) ValidateConfig() error
+```
+
+The implementation of this would check that the license key corresponds to an
+actual NewRelic account, perhaps by hitting some 'ping' endpoint you might have
+available, or by starting a demo transaction.
+
+We wouldn't want to add it to the Application interface in a minor version
+bump, because the interface is used as a return value from `NewApplication`
+(rather than a concrete struct).
+It could affect clients or their tests that use it.
+
+## Proposal 2
+
+Could this be checked as part of `newAppRun()`?
+
+### PR vs. Issues
+
+I'm filing this as a PR, but I'd rather have filed an issue. Is there a reason
+issues aren't enabled for your repo?


crenshaw:
Thanks for your proposal.  It's a bit wonky that the issues tab is closed on this repo.  New Relic's esteemed and talented support staff are best equipped at handling issues.  As the README.md states,

> Any questions or issues should be directed to our support site or our community forum.

Regarding your proposal, the Go Agent already does a couple of things to help customers discern whether they are attempting to connect with an invalid key.  First, the API call returns a non-nil error when the application cannot connect:

See: https://github.com/newrelic/go-agent/blob/462a146a1270954750aac98c5b3e83dec6a9adfd/application.go#L55-L63

Second, the Go Agent logs a couple of messages indicating that an invalid license key prevented the application from connecting.  For example:

```
go run main.go 
...
(1998) 2018/07/12 16:35:26.520859 {"level":"debug","msg":"rpm failure","context":{"command":"preconnect","error":"NewRelic::Agent::LicenseException: Invalid license key, please contact support@newrelic.com","url":"https://collector.newrelic.com/agent_listener/invoke_raw_method?license_key=2...2222\u0026marshal_format=json\u0026method=preconnect\u0026protocol_version=16"}}
(1998) 2018/07/12 16:35:26.521162 {"level":"debug","msg":"rpm response","context":{"command":"preconnect","response":null,"url":"https://collector.newrelic.com/agent_listener/invoke_raw_method?license_key=2...22222\u0026marshal_format=json\u0026method=preconnect\u0026protocol_version=16"}}
```

It's hard to do better than the error and the logging; the Go Agent doesn't know whether the license is valid until some background routine finds out.

purple4reina:
Hi @cjcjameson,

We've recently published some proposed changes to the agent for a 3.0 major version release https://github.com/newrelic/go-agent/issues/106. We're wondering what you think of these changes and if they will fit your needs.

Particularly, we've proposed some changes around how an Application would be configured. What do you think of this? Do you think it will fit your needs? If not, what do you think could improve it?

PR #75, opened 2018-07-31
Allow segments to be started at an explicit time

robladbrook:
I'm working on a case where we handle a `post-execution` event which contains the start time of the operation in question but does not allow any means of starting the segment timing before the operation begins.

This PR introduces the ability to start a segment at a specific time which allows this use case to be handled.

### DIFF
--- a/internal_test.go
+++ b/internal_test.go
@@ -776,6 +776,8 @@ func TestZeroSegmentsSafe(t *testing.T) {
 
 	StartSegmentNow(nil)
 
+	StartSegmentAt(nil, time.Now())
+
 	ds := DatastoreSegment{}
 	ds.End()
 
--- a/internal_txn.go
+++ b/internal_txn.go
@@ -453,10 +453,14 @@ func (txn *txn) Ignore() error {
 }
 
 func (txn *txn) StartSegmentNow() SegmentStartTime {
+	return txn.StartSegmentAt(time.Now())
+}
+
+func (txn *txn) StartSegmentAt(startedAt time.Time) SegmentStartTime {
 	var s internal.SegmentStartTime
 	txn.Lock()
 	if !txn.finished {
-		s = internal.StartSegment(&txn.TxnData, time.Now())
+		s = internal.StartSegment(&txn.TxnData, startedAt)
 	}
 	txn.Unlock()
 	return SegmentStartTime{
--- a/segments.go
+++ b/segments.go
@@ -1,6 +1,9 @@
 package newrelic
 
-import "net/http"
+import (
+	"net/http"
+	"time"
+)
 
 // SegmentStartTime is created by Transaction.StartSegmentNow and marks the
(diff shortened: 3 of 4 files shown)


astormnewrelic:
@robladbrook,

Thank you for sharing your customization via a PR -- its always exciting to see what folks are doing with out agent(s) out in the world!

That said, we're not sure we're _quite_ ready to expose a public API method for setting arbitrary segment start times -- this has some implications for how we calculate exclusive time that could be problematic for some of our users.  

If you have a moment (totally optional) -- would you be able to share a few more details of your use case?  When other users have tackled similar problems they usually turn to [custom attributes](https://docs.newrelic.com/docs/agents/go-agent/instrumentation/go-agent-attributes).  Would custom attributes have solved your problem, or is there some nuance here that made overriding the segment start time a better solution for you?

robladbrook:
The specific example I'm using this for is timing [go-pg](https://github.com/go-pg/pg) queries. The reason why custom attributes won't work is because we can't achieve all the proper reporting and timing that you get as part of datastore segments.

The hook I'm using is: [OnQueryProcessed](https://github.com/go-pg/pg/blob/583144043717b73fa07a49b6182a722a0452181f/hook.go#L69)

It's means instead of writing code like this:
```go
func GetCountry(iso string, conn orm.DB, nrTxn newrelic.Transaction) (Country, error) {
  s := newrelic.DatastoreSegment{Product: newrelic.DatastorePostgres, Collection: "countries", Operation: "SELECT"}
  s.StartTime = newrelic.StartSegmentNow(nrTxn)
  var country Country
  err := conn.Model(&country).
    Where("iso_alpha2_code = ?", iso).
    Select()
  s.End()
  return country, err
}
```

I can instead write code like this:

```go
func GetCountry(iso string, conn orm.DB) (Country, error) {
  var country Country
  err := conn.Model(&country).
    Where("iso_alpha2_code = ?", iso).
    Select()
  return country, err
}
```

astormnewrelic:
Thanks again Rob, that's useful feedback, and I can see why a custom attribute + custom insights reporting isn't an ideal fit for your situation.

While the aforementioned issues around timing calculations for _all_ our customers prevent us from accepting this PR at this time, we will (thanks to your PR) be thinking harder about these problems moving forward.  Both the specific idea of allowing end-user-programmers to dictate the start time on a segment, as well as the general idea of lighter weight instrumentation of code (i.e. your example of not needing an explicit datastore segment when instrumenting your `go-pg` queries).

Thank you again for making the PR, and please don't hesitate to make PRs in the future, or to bring these issues up [on our forums](http://forum.newrelic.com/) or [via support](http://support.newrelic.com/).

tompave:
Hello @astormnewrelic, do you have any update on this?

astormnewrelic:
@tompave Same status [as from August 7th](https://github.com/newrelic/go-agent/pull/75#issuecomment-411220624).

PR #76, opened 2018-11-06
Applied gofmt on the repo

som-poddar:
Formatted the code by applying [gofmt](https://golang.org/cmd/gofmt/).

### DIFF
--- a/internal/cat_test.go
+++ b/internal/cat_test.go
@@ -100,7 +100,7 @@ func TestCatMap(t *testing.T) {
 		tr.CrossProcess.Init(true, false, &ConnectReply{
 			CrossProcessID:  "1#1",
 			EncodingKey:     "foo",
-			TrustedAccounts: map[int]struct{}{1: struct{}{}},
+			TrustedAccounts: map[int]struct{}{1: {}},
 		}, CrossProcessMetadata{})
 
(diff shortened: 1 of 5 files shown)


crenshaw:
Thanks for the PR @som-poddar!  Can you let us know what flags, additional ruleset, or go  versions of you're running with `gofmt`?  We're pretty strict about applying gofmt ourselves, so we're interested in what extra steps you're taking (if any) to get the results you submitted.  Also, FWIW, your changes don't seem to line up with the formatting on the official Go Playground  https://play.golang.org

Regarding the issues tab, right now the best way to submit an issue is via customer support.  This ensures the issue's tracked for all our customers, and that those customers (including you!) get resolutions for known issues as quickly as possible.

Also, as you've noticed, we do accept PRs to the repo here which can serve as a an informal way to let us know about issues.  But you didn't hear me say that, and I suspect someone may ask me to take this out of my draft so I better reply to you quickly :)

Thanks again for the PR, looking forward to hearing your answers!

som-poddar:
@crenshaw, I used the following.
```
gofmt -s -w .
```

willnewrelic:
Hi @som-poddar 

Thanks for bringing this to our attention.  We will be sure to use `gofmt -s` prior to the next release.

PR #77, opened 2018-12-01
add in Start/End Backdated funcs for Segments

cgilling:
I was looking into how I could handle more automatic reporting about database transactions with Cassandra, and the gocql package provides an easy way to hook into every query being executed using something called a [QueryObserver](https://godoc.org/github.com/gocql/gocql#QueryObserver). This gets called after the query is finished with the Start and End times given in the provided object.

In order to use this, I need a way to be able to set segment start and end times to those other than `time.Now()`, hence these changes I made. It didn't seem like this should really be a problem given the model of how segments are only supposed to belong to a single non-concurrent transaction, so allowing for a backdated segment that fits within the model of segments being a stack doesn't cause any problems.

Looking at the past pull requests it looks like they are often taken internally and applied to an internal repo if they are acceptable, so I didn't bother to add as much documentation/tests as ai might normally in a pull request, I hope that is okay. Another thing I was considering is whether the `StartBackdatedSegment` function should return and error if the start time is too early. I left that out for now, but seems like it might be good to add in.

### DIFF
--- a/internal_txn.go
+++ b/internal_txn.go
@@ -529,12 +529,27 @@ func (txn *txn) StartSegmentNow() SegmentStartTime {
 	}
 }
 
+func (txn *txn) StartBackdatedSegment(t time.Time) SegmentStartTime {
+	var s internal.SegmentStartTime
+	txn.Lock()
+	if !txn.finished {
+		s = internal.StartSegment(&txn.TxnData, t)
+	}
+	txn.Unlock()
+	return SegmentStartTime{
+		segment: segment{
+			start: s,
+			txn:   txn,
+		},
+	}
+}
+
 type segment struct {
 	start internal.SegmentStartTime
 	txn   *txn
 }
 
-func endSegment(s *Segment) error {
+func endSegment(s *Segment, t time.Time) error {
 	txn := s.StartTime.txn
 	if nil == txn {
 		return nil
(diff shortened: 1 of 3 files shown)


cgilling:
Also here is a quick example of some code that would use it:
```go
type CQLQueryObserver struct {
}

func (qo *CQLQueryObserver) ObserveQuery(ctx context.Context, query gocql.ObservedQuery) {
	txn := newrelic.FromContext(ctx)
	if txn == nil {
		return
	}
	splitQuery := strings.Fields(query.Statement)
	s := newrelic.DatastoreSegment{
		StartTime:          txn.StartBackdatedSegment(query.Start),
		Product:            newrelic.DatastoreCassandra,
		ParameterizedQuery: query.Statement,
		DatabaseName:       query.Keyspace,
	}
	if len(splitQuery) >= 1 {
		s.Operation = splitQuery[0]
	}
	for i, part := range splitQuery {
		if part == "FROM" || part == "from" {
			if i+1 < len(splitQuery) {
				s.Collection = splitQuery[i+1]
			}
			break
		}
	}
	s.BackdatedEnd(query.End)
}```

willnewrelic:
Hi cgilling!

Thanks for creating this PR!

We are very intrigued by the idea:  Using the observer callbacks seems like a much cleaner solution than wrapping `gocql.Batch` and `gocql.Query`.

This could interfere with our metrics if the start or end times overlap with previously created segments.  So if we included this functionality, we would likely only create the segment if it started after the most recently ended segment.  I think that's fine though:  No one should be creating other segments during the Cassandra call, right?

I can't guarantee this will make our roadmap or be released, but I assure you that we are think about this.  Stay tuned!

PR #84, opened 2019-02-15
Improved Async Support:  Transaction.NewGoroutine

willnewrelic:
Hi All!

We are excited to show you our plans for improved async support!  This PR introduces a new `Transaction.NewGoroutine() Transaction` method which allows transactions to create segments in multiple goroutines!

`NewGoroutine` returns a new reference to the `Transaction`.  This must be called any time you are passing the `Transaction` to another goroutine which makes segments.  Each segment-creating goroutine must have its own `Transaction` reference.  It does not matter if you call this before or after the other goroutine has started.

All `Transaction` methods can be used in any `Transaction` reference.  The `Transaction` will end when `End()` is called in any goroutine.

Example passing a new `Transaction` reference directly to another goroutine:

```go
	go func(txn newrelic.Transaction) {
		defer newrelic.StartSegment(txn, "async").End()
		time.Sleep(100 * time.Millisecond)
	}(txn.NewGoroutine())
```

Example passing a new `Transaction` reference on a channel to another
goroutine:

```go
	ch := make(chan newrelic.Transaction)
	go func() {
		txn := <-ch
		defer newrelic.StartSegment(txn, "async").End()
		time.Sleep(100 * time.Millisecond)
	}()
	ch <- txn.NewGoroutine()
```

We would love your feedback!  Please let us know what you think!  Here are some particular questions we have:

* Ease of API use: Does calling `txn.NewGoroutine()` for each goroutine work for you? Does it function as you’d expect it to do? Is there any change you can suggest to the API that would make it easier for you to use?
* APM UI data. For each chart/graph, does the data displayed explain what is going on in your application? Will it be useful to you in uncovering performance problems?
  * Transaction Traces (summary and details)
  * App server breakdown graph

screen shot 2019-02-15 at 11 35 55 am

  * Breakdown table

screen shot 2019-02-15 at 11 36 21 am

This is a Beta release:

WARNING: THIS CODE IS IN DEVELOPMENT.  IT IS NOT A FINAL NEW RELIC PRODUCT OR
OFFERING. DO NOT USE IN PRODUCTION. USE AT YOUR OWN RISK. TERMS AND CONDITIONS
APPLY - SEE LICENSE FILE.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,37 @@
 ## ChangeLog
 
-* Introduced `Transaction.NewGoroutine() Transaction` method which allows the
-  transaction to be used in multiple goroutines!  `NewGoroutine` must be called
-  when the `Transaction` is passed to a different goroutine.  The entire
-  `Transaction` will end when `End()` is called in any goroutine.  Example use:
+* Introduced `Transaction.NewGoroutine() Transaction` method which allows
+  transactions to create segments in multiple goroutines!
+
+`NewGoroutine` returns a new reference to the `Transaction`.  This must be
+called any time you are passing the `Transaction` to another goroutine which
+makes segments.  Each segment-creating goroutine must have its own `Transaction`
+reference.  It does not matter if you call this before or after the other
+goroutine has started.
+
+All `Transaction` methods can be used in any `Transaction` reference.  The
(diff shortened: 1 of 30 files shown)


willnewrelic:
Hi all

We are pleased to announce that this functionality has been merged to master!  Enjoy!

PR #91, opened 2019-06-01
Clone request in RoundTripper to respect the interface contract

jlordiales:
The public docs for the `http.RoundTripper` interface specifies that `RoundTrip should not modify the request` (https://golang.org/pkg/net/http/#RoundTripper).

You can see an example of how the github client does a very similar thing here: https://github.com/google/go-github/blob/master/github/github.go#L955


jlordiales:
ping @willnewrelic

willnewrelic:
Hi @jlordiales  

Thanks for the PR, we appreciate the contribution.

Expect to see this fixed in our next release.  I will keep this PR open until then.

willnewrelic:
Hi @jlordiales

This fix is now released as part of 2.8.0!  Thanks again for the contribution.

Issue #100, opened 2019-09-02
Stacktrace frame lookup accuracy


lggomezml:
Hi

I'm working on a custom gin recover middleware that prints the offending stacktrace whenever a panic goes uncatched. In this scenario, we noticed an application's code path in which the line number on the printed stacktrace is wrong, returning line 59 instead of line 61 (it took me a long time to realize that something was wrong with the stacktrace itself and not the reported line):

```go
func (ls LocationsMutableService) getChildren(ctx context.Context, tree string, loc string) (LocationChildren, error) {
	url := ""
	var children LocationChildren

//line 59 - panic not possible here
	response, err := ls.Client.GET(url, &children)

//line 61 - actual place of panic (response == nil)
	if response.StatusCode == http.StatusNotFound { 
		return nil, errors.New("")
	}

	if err != nil {
		return nil, err
	}

	return children, err
}
```

We fixed this by writing a implementation based on debug.Stack() but, surprisingly, after deploying the fix to verify it I'm observing that the internal stacktrace used by txn.NoticeError has the same inaccuracy that our original middleware implementation had:

```
2019/08/30 14:33:09 /go/src/github.com/.../src/api/vendor/github.com/.../gingonic/mlhandlers/nr_stack.go 19 mlhandlers.GetStackTrace
/go/src/github.com/.../src/api/vendor/github.com/.../gingonic/mlhandlers/recovery.go 58 mlhandlers.RecoveryWithWriter.func1.1
/usr/local/go/src/runtime/panic.go 522 runtime.gopanic
/usr/local/go/src/runtime/panic.go 82 runtime.panicmem
/usr/local/go/src/runtime/signal_unix.go 390 runtime.sigpanic
/go/src/github.com/.../src/api/gateways/locations_mutable_service.go 59 gateways.LocationsMutableService.getChildren
/go/src/github.com/.../src/api/gateways/locations_mutable_service.go 51 gateways.LocationsMutableService.GetChildren.func1
```

After some research the main culprit seems to be a wrong PC being used when retrieving the stack frames, but I'm still not certain about the concrete cause. After checking the agent code I come across the frame lookup method which has the following comment about PC manipulation: 

https://github.com/newrelic/go-agent/blob/6730f18e627e631233292a59d8b9786e44f7ffb4/internal/stacktrace.go#L67-L76

As an experiment, I removed the substraction and it fixed the wrong line, although it messes the rest of the stack frames:

```
[...]

PR #101, opened 2019-09-05
Add stack frame lookup based on runtime.CallersFrames for supported g…

lggomez:
Part of the work required for #100 

Does not remove the dependency on runtime.Callers but it simplifies the implementation and removes the need for FuncForPC on Go 1.7+

### DIFF
--- a/internal/stackframes.go
+++ b/internal/stackframes.go
@@ -1,0 +1,36 @@
+// +build !go1.7
+
+package internal
+
+import (
+	"runtime"
+)
+
+func (st StackTrace) frames() []stacktraceFrame {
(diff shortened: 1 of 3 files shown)


newrelic-eheinlein:
@lggomez Thank you for this work!  It is coincidental timing, as your PR came in just as we finished up reviewing a PR internally that is very similar.  That PR is about to be merged internally.

One of my colleagues caught a bug in my code,  which I think is also present in your code, where the last frame of the stack trace is missed because when `cf.Next()` returns the final frame of the stack trace, `more` will be false, but the frame returned with it will still be valid.  

We will be merging our internal PR shortly, and the fix will be included in the next agent release.

PR #111, opened 2019-10-24
Using strings.EqualFold() for efficient string comparison

AdityaHarindar:
Improved efficiency compared to previous implementation, which uses `strings.ToLower()`

### DIFF
--- a/_integrations/nrawssdk/internal/internal.go
+++ b/_integrations/nrawssdk/internal/internal.go
@@ -65,7 +65,7 @@ func StartSegment(input StartSegmentInputs) *http.Request {
 
 	var segment endable
 	// Service name capitalization is different for v1 and v2.
-	if strings.ToLower(input.ServiceName) == "dynamodb" {
+	if strings.EqualFold(input.ServiceName, "dynamodb") {
 		segment = &newrelic.DatastoreSegment{
 			Product:            newrelic.DatastoreDynamoDB,
 			Collection:         getTableName(input.Params),


vivekv96:
@AdityaHarindar, this is slower when `input.ServiceName` is equal to `dynamodb`.

purple4reina:
@AdityaHarindar Thanks so much for your pull request. We always appreciate users helping make our code better! I'm working on reviewing it now and will get back to you shortly.

AdityaHarindar:
> @AdityaHarindar, this is slower when `input.ServiceName` is equal to `dynamodb`.

I’ve verified that this optimisation is faster in all but that case.

purple4reina:
Here is the benchmark file I've been using for testing:

```go
package benchy

import (
	"strings"
	"testing"
)

var (
	tolower   bool
	equalfold bool
)

func BenchmarkToLowerDifferent(b *testing.B) {
	var res bool
	for n := 0; n < b.N; n++ {
		res = strings.ToLower("Hello World") == "hello world"
	}
	tolower = res
}

func BenchmarkEqualFoldDifferent(b *testing.B) {
	var res bool
	for n := 0; n < b.N; n++ {
		res = strings.EqualFold("Hello World", "hello world")
	}
	equalfold = res
}

func BenchmarkToLowerSame(b *testing.B) {
	var res bool
	for n := 0; n < b.N; n++ {
		res = strings.ToLower("hello world") == "hello world"
	}
	tolower = res
}

func BenchmarkEqualFoldSame(b *testing.B) {
	var res bool
	for n := 0; n < b.N; n++ {
		res = strings.EqualFold("hello world", "hello world")
	}
	equalfold = res
}
```

Here's my results:

```
goos: darwin             
goarch: amd64            
BenchmarkToLowerDifferent-8             183815722               72.2 ns/op            16 B/op          1 allocs/op
BenchmarkEqualFoldDifferent-8           302717677               36.0 ns/op             0 B/op          0 allocs/op
BenchmarkToLowerSame-8                  591520122               20.2 ns/op             0 B/op          0 allocs/op
BenchmarkEqualFoldSame-8                391565965               29.5 ns/op             0 B/op          0 allocs/op
```

You are correct, this change is more efficient and reduces allocations when the compared strings are different but adds some overhead when the strings are the same.

I poked around at the value of `input.ServiceName` for aws-sdk v1 versus v2. In v1 the value is `"dynamodb"`. In v2 the value is `"DynamoDB"`. Therefore, this would be an optimization for v2 but add a bit of overhead for v1.

I will bring these results to my team and we will discuss what we would like to do. Again, we really appreciate you doing this work and making this contribution! We'll get back to you soon.

purple4reina:
Okay okay. We've got an even better idea!

```go
if input.ServiceName == "dynamodb" || input.ServiceName == "DynamoDB" {
```

And we already have tests that cover if AWS decides to change the casing for `ServiceName`. This solution benchmarks down to `0.284 ns/op`!!!!

purple4reina:
@AdityaHarindar, the optimization described above will go out with our next agent release!

Thanks again for bringing this to our attention. We are always happy to help out.

I will close this pull request. If you need anything else, you are always welcome to respond here and we will respond as soon as we can.

purple4reina:
@AdityaHarindar We just released v2.15.0 with the change explained above. Let us know what you think!

PR #113, opened 2019-10-28
Add txn to req context on nrhttprouter

flaviotruzzi:
The transaction is not being pushed into the Request Context.

This would make segments not work when using the recommended way of fetching the transaction from the context.


adimarco:
Is there a recommended way to do this while we wait for the PR to merge.  This basically breaks all NR integration when using `httprouter` in the documented way.

willnewrelic:
Hi @adimarco !  We expect to release later this week.

willnewrelic:
Hi all,

We have just released a fix for this in version 2.16.0!  Thanks for bringing this to our attention!

PR #118, opened 2019-11-24
Correct nrgorilla's description

edmorley:
Since it currently refers to gin rather than mux - which is presumably a leftover from when the READMEs were mass-added in 7664285.


newrelic-eheinlein:
Thank you for catching and correcting this mistake! We just did a release on Friday, so we'll incorporate this into our next release, which we expect to me a major release of version 3.0.

newrelic-eheinlein:
Hi @edmorley - actually, we were working internally on rearranging some of the integration files today in preparation for the 3.0 release, and so we have integrated the change you suggested into one of our internal pull requests. Your change will be in the next release. Thank you again for finding and correcting this!

edmorley:
Thank you :-)

PR #123, opened 2019-12-15
Return a noop gRPC server interceptors rather than nil

rogchap:
The gRPC interceptors currently return `nil`, however this can cause a panic when used in a chain of interceptors. Eg:
```go
return grpc.NewServer(grpc_middleware.WithUnaryServerChain(
	nrgrpc.UnaryServerInterceptor(nr), // <- this will panic if nr is nil
	grpc_ctxtags.UnaryServerInterceptor(),
	grpc_zap.UnaryServerInterceptor(logger, opts...),
	// etc...
))
```

This PR changes the behavior to return a noop server interceptors to avoid panics.

### DIFF
--- a/_integrations/nrgrpc/nrgrpc_server.go
+++ b/_integrations/nrgrpc/nrgrpc_server.go
@@ -59,7 +59,9 @@ func startTransaction(ctx context.Context, app newrelic.Application, fullMethod 
 //
 func UnaryServerInterceptor(app newrelic.Application) grpc.UnaryServerInterceptor {
 	if nil == app {
-		return nil
+		return func(ctx, context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+			return handler(ctx, req)
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


purple4reina:
@rogchap It indeed looks like you have found an unfortunate issue with our gRPC integration! Thanks for catching it! We'll include the fix in an upcoming release!

purple4reina:
The agent has been released with this update! Thanks again!

rogchap:
@purple4reina Thanks for getting this implemented; however it's very frustrating that this non-breaking change has gone into v3 and not back ported to v2 (as per this PR); 
Upgrading to v3 is a huge undertaking for our services due to the sheer number of breaking changes v3 introduces.

purple4reina:
Hey @rogchap,

We did in fact make this change in the v2 agent as well. Check it out, your name is even on the commit 😄 

https://github.com/newrelic/go-agent/blame/master/_integrations/nrgrpc/nrgrpc_server.go#L62-L63

I'm curious how was it you got the wrong impression? Maybe the confusion came from the change log which only mentioned v3? I'd love to make this experience better for you in the future.

Sorry again and take care,
Rey

PR #131, opened 2020-02-04
Update CHANGELOG.md

jodeev:
Need to document the known issue WRT .NET agent creating the TraceID

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,11 @@
 
   http.ListenAndServe(":8000", r)
   ```
+  
+### Known Issues and Workarounds
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -39,6 +39,11 @@
 
   http.ListenAndServe(":8000", r)
   ```
+  
+### Known Issues and Workarounds
+
+* If a .NET agent is initiating traces as the root service, do not upgrade your downstream
```
newrelic-eheinlein: Is it worth specifying that this is for **distributed** traces? (I assume customers would understand either way, but I feel like there are multiple contexts for the word "traces".)

purple4reina:
Thank you for your submission. We will deal with this internally then include it with our next release!

Issue #132, opened 2020-02-07
Integration accesses internal packages.


nicolas-martin:
Hi, I'd like to contribute and add a snowflake integration. However, since I forked the repo, I am not able to compile my module as it requires access to packages inside `/internal`.

In order to get this working, I had to clone the repo, create a new branch with my changes and use a local replace in my app. But isn't really viable in a production environment.

I plan on creating a PR and submitting my changes soon but ideally I would be able to use it myself without having to wait for the PR, merge and release of the go-agent package.

purple4reina:
Hey @nicolas-martin, thanks for submitting this issue! Which parts of the `internal` package in particular are you looking to have access to?

nicolas-martin:
My changes were heavily inspired by the [mysql integration](https://github.com/newrelic/go-agent/blob/master/v3/integrations/nrmysql/nrmysql.go). I need `github.com/newrelic/go-agent/v3/internal`, `github.com/newrelic/go-agent/v3/internal/sqlparse`

purple4reina:
Cool, that's what I suspected. Let me bring this to the team and see what we can do to help you out. In the meantime, thanks a ton for working on a Snowflake integration! I look forward to seeing it!

nicolas-martin:
Yup, I was able to test it in our newrelic instance. We don't get as much details as mysql does, but I do get query names and timing. Which is good enough for now.

purple4reina:
@nicolas-martin,

An update for you. We've decided to go ahead and move the `ParseQuery` code out of the `internal` package! 🎉 It'll be available in our next release.

As for your snowflake integration and use of the `ParseQuery` code (which for the record I have not looked at your PR yet), my one concern about you using the `ParseQuery` would be for security reasons. It is important that we do not send any more than table names and column names. I would be concerned that because snowflake is not exactly the same language as mysql, postgres, or sqlite, that our regexes might behave in strange ways. Anyway, that's what I'll be looking at as I review that PR, but I wanted to include it here too so others reading this can be aware.

More soon,
Rey

purple4reina:
@nicolas-martin, we just released a version of the go-agent with the `ParseQuery` code made public. Thanks again for your suggestion and contributions!

nicolas-martin:
Thank you for the quick turn around!

purple4reina:
@nicolas-martin,

[...]

PR #133, opened 2020-02-07
Snowflake integration

nicolas-martin:
Initial implementation of a snowflake integration. I was able to verify that this is working, however the details aren't as fine grained as other sql integration. I was able to see, the name, table and instance location but not the full query. I think that would need more changes which I'm not too familiar with. However this is good enough for a general use.

re: #132. An alternative to waiting on the full release cycle would be merging this into a separate branch which would allow me to fetch this change using `go get github.com/newrelic/go-agent/v3/integrations/nrsnowflake@branchname`


purple4reina:
@nicolas-martin Thanks so much for this PR! This looks great. I'll take a deeper look early next week and get back to you.

purple4reina:
@nicolas-martin This all looks great! We'll get this code out in our next release. Thanks a ton for adding it. We're happy to be able to support Snowflake!

purple4reina:
@nicolas-martin, thanks again for your contributions. Support for Snowflake has just been released in the `nrsnowflake` package!

PR #135, opened 2020-02-22
Use c.FullPath() to name Gin transactions

aymericbeaumet:
The `c.FullPath()` function has recently been added to the Gin API (https://godoc.org/github.com/gin-gonic/gin#Context.FullPath).

It allows for more expressive transaction naming, and also prevent to use reflection in the hot path.

It changes the current behavior for existing users, so we might want to make this available behind some flag until the next major release.

### DIFF
--- a/v3/integrations/nrgin/nrgin.go
+++ b/v3/integrations/nrgin/nrgin.go
@@ -100,7 +100,7 @@ func Transaction(c Context) *newrelic.Transaction {
 func Middleware(app *newrelic.Application) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		if app != nil {
-			name := c.Request.Method + " " + c.HandlerName()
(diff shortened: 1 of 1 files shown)


purple4reina:
@aymericbeaumet Thanks a ton for this pull request. We think updating the transaction name based on full path would be a much needed improvement.

We're going to need the integration package to still support older versions of Gin as well plus offer a friendly upgrade path. I'll see what I can do for those two things and get back to you.

aymericbeaumet:
@purple4reina Great! I've updated the PR to deal with the edge-case of an empty full-path.

I'm not sure how to proceed for the older Gin versions. But regarding the friendly upgrade path, we can create a different function `MiddlewareFullPath` and keep the current one pristine.

purple4reina:
@aymericbeaumet Thanks a ton for your contribution! This work has now been released in v3.4.0!!

PR #163, opened 2020-06-15
Improve experience when sending data times out.

purple4reina:
This pre-release branch does two things to improve the agent experience when sending data to the New Relic backend times out.

First, instead of dropping data on the floor, we will save it and attempt to resend it with the next harvest cycle. Note that because the agent has internal limits to the amount of data it holds, if the timeouts happen for a prolonged period and your application is collecting a lot of events, some could still get dropped.

Second, the agent will stop using the `http.DefaultTransport` for communication with the New Relic backend. This change will help address https://github.com/golang/go/issues/33006 which can cause all outbound requests using the same transport to hang, irregardless of the host they are attempting to connect with.

### DIFF
--- a/v3/newrelic/collector.go
+++ b/v3/newrelic/collector.go
@@ -74,6 +74,8 @@ type rpmResponse struct {
 	// should be used to avoid mismatch between statusCode and Err.
 	Err                      error
 	disconnectSecurityPolicy bool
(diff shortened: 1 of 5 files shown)


yuanmwang-wf:
Looks good to me 👍

Issue #138, opened 2020-03-21
orphaned spans with nrmysql integration


amsipe:
Currently we are seeing issues with orphaned spans when using the `nrmysql` integration and the [sqlx](https://github.com/jmoiron/sqlx) db wrapper. Any thoughts as to what might be causing these spans to get misaligned?

purple4reina:
Thanks a ton for opening this issue @amsipe!

As I stated in issue #139, there are a couple of reasons why you might see orphaned spans in the Distributed Tracing UI:

* Ending a segment after the transaction has ended. This can be a commonly unnoticed occurrence when using the `txn.NewGoroutine` api.
* Forgetting to call `End()` on a running segment.

Either of these situations would mean that a span was started in the agent but never sent to the backend. So, when the UI tries to render the trace, there's a missing parent and the children go to the orphanage. 😢 

Before you opened this issue, we determined together that it is our sql instrumentation that is causing the missing spans. https://github.com/newrelic/go-agent/blob/b9b65ab3d35fb31361c9869fa0788cdb60dc0f4e/v3/newrelic/sql_driver.go#L163-L168

Unfortunately, we only conditionally end segments in some cases and the `nrmysql` driver happens to hit one of these cases.

I have gone ahead and filed this as a bug internally. We will try to get a fix out for it soon. At this point I'm thinking something along the lines of https://github.com/newrelic/go-agent/pull/75 but that is still open for discussion.

Thanks again. While I don't like releasing bugs, it does make me really happy when we can discover and fix them together.

Rey

purple4reina:
Hey @amsipe ! We just released v3.8.1 of the agent with a fix for this issue! Try it out and let us know how it works for you!

PR #167, opened 2020-07-02
Fix parentage of datastore segments

newrelic-eheinlein:
Addressing https://github.com/newrelic/go-agent/issues/138

If a SQL driver returned `driver.ErrSkip`, we were ending up with with orphaned datastore spans. This change conditionally creates the spans only if `driver.ErrSkip` is not returned.


newrelic-eheinlein:
A note on the failing Snyk check: we need to get the correct key added as a secret, which will take some coordination, but in the meantime this PR did not change any dependencies so we are not worried about it.

PR #176, opened 2020-07-11
Remove Nil Test Contexts

alrs:
The docs for `context` admonish:
> Do not pass a nil Context, even if a function permits it. Pass context.TODO if you are unsure about which Context to use. 

This replaces nil context variables that were in the tests with `context.Background()`.


newrelic-eheinlein:
Thank you for this PR! We will review it shortly and let you know if we have any feedback.

PR #187, opened 2020-07-21
Add Segment and NewGoroutine support.

purple4reina:
This PR adds support for basic `Segment`s and `Transaction.NewGoroutine`.

Not included is support for other segment types.

### DIFF
--- a/v4/newrelic/segments.go
+++ b/v4/newrelic/segments.go
@@ -4,13 +4,28 @@
 package newrelic
 
 import (
+	"context"
 	"net/http"
+
+	"go.opentelemetry.io/otel/api/trace"
 )
+
+type span struct {
+	Span trace.Span
+	ctx  context.Context
+	// TODO: linked list and ballooning memory?
+	parent *span
+	ended  bool
+	// TODO: reference cycles?
+	txn *Transaction
+}
 
 // SegmentStartTime is created by Transaction.StartSegmentNow and marks the
 // beginning of a segment.  A segment with a zero-valued SegmentStartTime may
 // safely be ended.
-type SegmentStartTime struct{}
+type SegmentStartTime struct {
+	*span
+}
 
 // Segment is used to instrument functions, methods, and blocks of code.  The
 // easiest way use Segment is the Transaction.StartSegment method.
@@ -126,6 +141,21 @@ const (
 	MessageExchange MessageDestinationType = "Exchange"
 )
 
+func (s *span) end() {
+	s.Span.End()
+	s.ended = true
+	if s.txn != nil {
+		parent := s.parent
+		for parent != nil {
+			if !parent.ended {
+				s.txn.setCurrentSpan(s.parent)
+				return
+			}
+			parent = parent.parent
+		}
+	}
+}
+
 // AddAttribute adds a key value pair to the current segment.
 //
 // The key must contain fewer than than 255 bytes.  The value must be a
@@ -133,7 +163,16 @@ const (
 func (s *Segment) AddAttribute(key string, val interface{}) {}
 
 // End finishes the segment.
-func (s *Segment) End() {}
+func (s *Segment) End() {
+	if s == nil {
+		return
+	}
+	if s.StartTime.span == nil {
+		return
+	}
+	s.StartTime.Span.SetName(s.Name)
+	s.StartTime.end()
+}
 
 // AddAttribute adds a key value pair to the current DatastoreSegment.
 //
--- a/v4/newrelic/segments.go
+++ b/v4/newrelic/segments.go
@@ -13,13 +13,11 @@ import (
 
 type span struct {
 	sync.Mutex
-	Span trace.Span
-	ctx  context.Context
-	// TODO: linked list and ballooning memory?
+	Span   trace.Span
+	ctx    context.Context
 	parent *span
 	ended  bool
-	// TODO: reference cycles?
-	txn *Transaction
+	thread *thread
 }
 
 // SegmentStartTime is created by Transaction.StartSegmentNow and marks the
@@ -148,15 +146,13 @@ func (s *span) end() {
 	s.Lock()
 	s.ended = true
 	s.Unlock()
-	if s.txn != nil {
-		parent := s.parent
-		for parent != nil {
-			if !parent.isEnded() {
-				s.txn.thread.setCurrentSpan(parent)
-				return
-			}
-			parent = parent.parent
+	parent := s.parent
+	for parent != nil {
+		if !parent.isEnded() {
+			s.thread.setCurrentSpan(parent)
+			return
 		}
+		parent = parent.parent
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pyohannes:
Should we extend `.github/workflows/ci.yaml` to include `v4` tests? Or is that not intended at this stage?

v4/newrelic/transaction.go
```diff
@@ -0,0 +1,348 @@
...
+// ExternalSegment.  The returned SegmentStartTime is safe to use even  when the
+// Transaction receiver is nil.  In this case, the segment will have no effect.
+func (txn *Transaction) StartSegmentNow() SegmentStartTime {
+	if txn == nil {
+		return SegmentStartTime{}
+	}
+	if txn.isEnded() {
+		return SegmentStartTime{}
+	}
+	parent := txn.thread.getCurrentSpan()
```
pyohannes, line 132: Do you think we could obtain the current span from OTel, without managing it in the shim layer? I think otherwise we might run into problems mixing the Go Agent and OTel instrumentation. I'm thinking about a test case like this:
```go
app, tracer := newTestApp(t)
txn := app.StartTransaction("transaction")

seg1 := txn.StartSegment("seg1")

_, span := tracer.Start(context.Background(), "seg2")
span.End()

seg1.End()
txn.End()
```
In the ideal case, this should link parent `seg2` to `seg1`. But I think managing the current span in the transaction state interferes with that.

pyohannes: I think we also should add something similar to the code snippet above as a unit test case.

v4/newrelic/segments_test.go
```diff
@@ -0,0 +1,174 @@
...
+	return s.SpanContext().SpanID.String()
+}
+
+func getParentID(s trace.Span) string {
+	return s.(*testtrace.Span).ParentSpanID().String()
+}
+
+func newTestApp(t *testing.T) *Application {
+	app, err := NewApplication(func(cfg *Config) {
+		cfg.OpenTelemetry.Tracer = testtrace.NewTracer()
```
pyohannes, line 23: That's a great way to test this!

v4/newrelic/transaction.go
purple4reina: Fantastic point. I will consider this.

pyohannes: I had a closer look at this, and I'm actually not sure how far we can get without managing parenting in the shim layer.

When looking at OpenTelemetry Go, I see that the parenting is handling by using context. The parent context is passed to `tracer.Start`, which in turn returns the span and an associated context. This can then be passed on to further `tracer.Start` calls to create children of the context passed in. The currently active span is contained in the context, and is used by `tracer.Start` to set the parent span.

Maybe we have to stay with the solution you came up with. We could then introduce a new `GetContext` method on the transaction (maybe there's a better name), which returns the context related to the currently active segment. The sample above would then look like this:
```go
app, tracer := newTestApp(t)
txn := app.StartTransaction("transaction")

seg1 := txn.StartSegment("seg1")

_, span := tracer.Start(txn.GetContext(), "seg2")
span.End()

seg1.End()
txn.End()
```

purple4reina: Yes, this is what I was thinking as well. However instead of a `GetContext` we could just use the already existing `NewContext` method.

```go
app, tracer := newTestApp(t)
txn := app.StartTransaction("transaction")

seg1 := txn.StartSegment("seg1")

_, span := tracer.Start(ctx, "seg2")
span.End()

seg1.End()
txn.End()
```

purple4reina: Hmmm, though you might be right about needing a new `GetContext` method. Either way, I think that decision can wait until we begin work on that aspect of the project.

v4/newrelic/segments.go
```diff
@@ -126,14 +143,46 @@ const (
 	MessageExchange MessageDestinationType = "Exchange"
 )
 
+func (s *span) end() {
+	s.Span.End()
+	s.Lock()
+	s.ended = true
+	s.Unlock()
+	if s.txn != nil {
```
pyohannes, line 151: Should we avoid doing this work when the segment was ended before?

I'm not sure what guarantees the Go agent makes for multiple calls to `segment.End`, but when multiple calls to `segment.End` are not caught, parenting might be off. For example in this case:
```go
txn := app.StartTransaction("transaction")
seg1 := txn.StartSegment("seg1")
seg2 := txn.StartSegment("seg2")
seg2.End()
seg1.End()
seg3 := txn.StartSegment("seg3")
seg2.End() // That's a weird one.
seg4 := txn.StartSegment("seg4")
seg4.End()
seg3.End()
txn.End()
```

I'm not sure to me whether the Go agent suggests that subsequent calls to `segment.End` have no effect. If not, feel free to ignore this comment!

v4/newrelic/segments.go
```diff
@@ -4,13 +4,30 @@
...
 
+type span struct {
+	sync.Mutex
+	Span trace.Span
+	ctx  context.Context
+	// TODO: linked list and ballooning memory?
+	parent *span
+	ended  bool
+	// TODO: reference cycles?
+	txn *Transaction
```
pyohannes, line 22: As far as I know, the Golang garbage collector should do fine with reference cycles. Cycles would be more of a problem for languages who do actual reference counting.

v4/newrelic/transaction.go
purple4reina: Hmm, and we should consider the case where the OTel span is the parent of a newrelic segment. I'm not sure how to handle that. Any ideas?

v4/newrelic/segments_test.go
```diff
@@ -0,0 +1,376 @@
...
+	seg1ParentID := getParentID(seg1.StartTime.Span)
+	seg2ParentID := getParentID(seg2.StartTime.Span)
+	spanParentID := getParentID(span)
+
+	if seg1ParentID != txnID {
+		t.Errorf("seg1 is not a child of txn: seg1ParentID=%s, txnID=%s",
+			seg1ParentID, txnID)
+	}
+	// XXX: Once NewContext has been implemented, the following test should
+	// fail.  span should be a child of seg1, but currently it has no parent.
```
pyohannes, line 326: Thinks. This already helps to lay out the context work!

v4/newrelic/transaction.go
```diff
@@ -125,7 +181,10 @@ func (txn *Transaction) StartSegmentNow() SegmentStartTime {
 //	// ... code you want to time here ...
 //	segment.End()
 func (txn *Transaction) StartSegment(name string) *Segment {
-	return nil
+	return &Segment{
+		StartTime: txn.StartSegmentNow(),
+		Name:      name,
```
pyohannes, line 186: Is there any reason to store the name instead of passing it right to `tracer.Start` via `StartSegmentNow`?

purple4reina: Yeah, good question. I did it this way because a customer could do something like this:

```go
seg := &Segment{
    StartTime: txn.StartSegmentNow(),
}
seg.Name = "hello"
seg.End()
```

In this example, they have set the name _after_ they have started it.

PR #189, opened 2020-07-27
Add v4 tests to actions.

purple4reina:
Adds v4 tests to the github actions workflow. Skipping integration tests for now because those ones certainly will fail at this time.

I removed two packages under the `internal` directory: `awssupport` and `integrationsupport`. Both of these are currently used by integrations only, and not the agent proper. I attempted to rip out most of their code, and leave them as just noop packages, but because they are so entwined with the agent as is, including requiring a `internal.ConnectReply` and `internal.HarvestTesting`, doing so was non-trivial. Instead, I opted to just remove them. We can put them back when we get to the point where we are looking to add testing and support for all our integration packages back.


v4/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/newrelic/go-agent/v4
 
-go 1.7
+go 1.13
```
purple4reina, line 3: OpenTelemetry-Go only supports go versions 1.13 and above: https://github.com/open-telemetry/opentelemetry-go/blob/master/.circleci/config.yml#L3-L10

build-script.sh
```diff
@@ -36,17 +36,21 @@ for dir in $DIRS; do
   if [ $dir == "." ]; then
     rm -rf v3/
     rm -rf v4/
-  else
+  elif [ $dir = v3* ]; then
```
pyohannes, line 39: :+1: for this.

PR #190, opened 2020-07-29
Implement other segment types.

purple4reina:
This pull request implements `DatastoreSegment`, `ExternalSegment`, and `MessageProducerSegment`. The biggest challenge here was deciding on the name for each of these span types. I did my best to follow the OpenTelemetry specs for naming.

* [Semantic conventions for database client calls](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md)
* [Semantic conventions for HTTP spans](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#name)
* [Semantic conventions for messaging systems](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#span-name)

### DIFF
--- a/v4/newrelic/segments.go
+++ b/v4/newrelic/segments.go
@@ -239,7 +239,14 @@ func (s *ExternalSegment) End() {
 }
 
 func (s *ExternalSegment) name() string {
-	return s.method() + ": " + s.host()
+	return s.library() + " " + s.method() + " " + s.host()
+}
+
+func (s *ExternalSegment) library() string {
+	if s.Library == "" {
+		return "http"
+	}
+	return s.Library
 }
 
 func (s *ExternalSegment) host() string {
--- a/v4/newrelic/segments.go
+++ b/v4/newrelic/segments.go
@@ -284,7 +284,27 @@ func (s *ExternalSegment) method() string {
 func (s *MessageProducerSegment) AddAttribute(key string, val interface{}) {}
 
 // End finishes the message segment.
-func (s *MessageProducerSegment) End() {}
+func (s *MessageProducerSegment) End() {
+	if s == nil {
+		return
+	}
+	if s.StartTime.span == nil {
+		return
+	}
+	if s.StartTime.isEnded() {
+		return
+	}
+	s.StartTime.Span.SetName(s.name())
+	s.StartTime.end()
+}
+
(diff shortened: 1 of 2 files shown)


v4/newrelic/segments.go
```diff
@@ -199,7 +224,58 @@ func (s *DatastoreSegment) End() {}
 func (s *ExternalSegment) AddAttribute(key string, val interface{}) {}
 
 // End finishes the external segment.
-func (s *ExternalSegment) End() {}
+func (s *ExternalSegment) End() {
+	if s == nil {
+		return
+	}
+	if s.StartTime.span == nil {
```
pyohannes, line 231: I wonder this check and everything below could be moved into `span.end`? So we just call `s.StartTime.end(s.name())` here? This would avoid code quadruplification. I'm not sure though whether that would be considered idiomatic Go.

v4/newrelic/segments.go
```diff
@@ -190,7 +191,31 @@ func (s *Segment) End() {
...
+		return
+	}
+	if s.StartTime.isEnded() {
+		return
+	}
+	s.StartTime.Span.SetName(s.name())
```
pyohannes, line 204: Will setting the `SpanKind` be part of this PR?

v4/newrelic/segments_test.go
```diff
@@ -374,3 +379,246 @@ func TestSegmentsOnEmptyTransaction(t *testing.T) {
...
+	}{
+		{
+			seg:  &ExternalSegment{},
+			name: "unknown: unknown",
+		},
+		{
+			seg: &ExternalSegment{
+				Host: "myhost:1234",
+			},
+			name: "unknown: myhost:1234",
```
pyohannes, line 505: Maybe we should add the value of `Library` at the beginning? So we'd get something like `HTTP GET www.newrelic.com`. That might be especially useful for non-HTTP external segments. That's also close to the semantic convention for HTTP spans, which suggest the minimal solution `"HTTP {METHOD_NAME}"`.

v4/newrelic/segments.go
pyohannes: I just saw the other PR, nevermind.

PR #191, opened 2020-07-29
Add DT support to the shim

pyohannes:
This adds DT support to the shim by implementing the functions `InsertDistributedTraceHeaders` and `AcceptDistributedTraceHeaders`, using the TraceContext injector and extractor OpenTelemetry provides.

The implementation for `InsertDistributedTraceHeaders` is straight-forward and uncontroversial.

### Problem

Things get a bit messy with `AcceptDistributedTraceHeaders` though. OpenTelemetry requires that the inbound headers were received _before_ the trace was started, whereas the Go agent requires the opposite, namely that a transaction was started _before_ inbound headers were received. We cannot call `AcceptDistributedTraceHeaders` before a transaction was started. I don't see a possibility to do a 1:1 mapping of one solution to another.

### Solution

The best solution I could come up with is starting a new trace after a call to `AcceptDistributedTraceHeaders`. A call to `AcceptDistributedTraceHeaders` creates a new context that will have the inbound trace id and the inbound span id, this context will then be used to start the next segment/span. With this approach, we'll end up with two (or more) traces for one transaction: one trace starting with the root transaction segment and now remote parent, another trace starting after a call to `AcceptDistributedTraceHeaders`, parented to the remote span and having the remote trace id.

Based on good feedback from @purple4reina I added special treatment for transactions for which no segments (besides the root segment) were started before the `AcceptDistributedTraceHeaders` call: in that case, during the call of `AcceptDistributedTraceHeaders` the root segment will be replaced with a new segment that is parented to the remote span context.

### DIFF
--- a/v4/newrelic/application.go
+++ b/v4/newrelic/application.go
@@ -30,6 +30,7 @@ func (app *Application) StartTransaction(name string) *Transaction {
 		thread: &thread{
 			currentSpan: s,
 		},
+		name: name,
 	}
 }
 
--- a/v4/newrelic/application.go
+++ b/v4/newrelic/application.go
@@ -8,13 +8,15 @@ import (
 	"time"
 
 	"go.opentelemetry.io/otel/api/global"
+	"go.opentelemetry.io/otel/api/propagation"
 	"go.opentelemetry.io/otel/api/trace"
 )
 
 // Application represents your application.  All methods on Application are nil
 // safe.  Therefore, a nil Application pointer can be safely used as a mock.
 type Application struct {
-	tracer trace.Tracer
+	tracer      trace.Tracer
+	propagators propagation.Propagators
 }
 
 // StartTransaction begins a Transaction with the given name.
@@ -30,6 +32,7 @@ func (app *Application) StartTransaction(name string) *Transaction {
 		thread: &thread{
 			currentSpan: s,
 		},
+		app: app,
 	}
 }
 
@@ -104,5 +107,9 @@ func NewApplication(opts ...ConfigOption) (*Application, error) {
 	if nil == tracer {
 		tracer = global.Tracer("traceName")
 	}
-	return &Application{tracer: tracer}, nil
+	propagators := propagation.New(
+		propagation.WithInjectors(trace.TraceContext{}),
+		propagation.WithExtractors(trace.TraceContext{}))
+
+	return &Application{tracer: tracer, propagators: propagators}, nil
 }
--- a/v4/newrelic/application.go
+++ b/v4/newrelic/application.go
@@ -30,7 +30,6 @@ func (app *Application) StartTransaction(name string) *Transaction {
 		thread: &thread{
 			currentSpan: s,
 		},
-		name: name,
 	}
 }
 
--- a/v4/newrelic/application.go
+++ b/v4/newrelic/application.go
@@ -30,8 +30,8 @@ func (app *Application) StartTransaction(name string) *Transaction {
 	return &Transaction{
 		rootSpan: s,
 		thread: &thread{
-			currentSpan: s,
-			spanCount:   1,
+			currentSpan:  s,
+			isSingleSpan: true,
 		},
 		app:  app,
 		name: name,
--- a/v4/newrelic/application.go
+++ b/v4/newrelic/application.go
@@ -31,8 +31,10 @@ func (app *Application) StartTransaction(name string) *Transaction {
 		rootSpan: s,
 		thread: &thread{
 			currentSpan: s,
+			spanCount:   1,
 		},
-		app: app,
+		app:  app,
+		name: name,
 	}
 }
 
--- a/v4/newrelic/application.go
+++ b/v4/newrelic/application.go
@@ -30,8 +30,7 @@ func (app *Application) StartTransaction(name string) *Transaction {
 	return &Transaction{
 		rootSpan: s,
 		thread: &thread{
-			currentSpan:  s,
-			isSingleSpan: true,
+			currentSpan: s,
 		},
 		app:  app,
 		name: name,
@@ -109,9 +108,11 @@ func NewApplication(opts ...ConfigOption) (*Application, error) {
 	if nil == tracer {
 		tracer = global.Tracer("traceName")
 	}
-	propagators := propagation.New(
-		propagation.WithInjectors(trace.TraceContext{}),
-		propagation.WithExtractors(trace.TraceContext{}))
-
+	propagators := c.OpenTelemetry.Propagators
+	if nil == propagators {
+		propagators = propagation.New(
+			propagation.WithInjectors(trace.TraceContext{}),
+			propagation.WithExtractors(trace.TraceContext{}))
+	}
 	return &Application{tracer: tracer, propagators: propagators}, nil
 }
--- a/v4/newrelic/config.go
+++ b/v4/newrelic/config.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"time"
 
+	"go.opentelemetry.io/otel/api/propagation"
 	"go.opentelemetry.io/otel/api/trace"
 )
 
@@ -340,6 +341,8 @@ type Config struct {
 	OpenTelemetry struct {
 		// Tracer TODO
 		Tracer trace.Tracer
+		// Propagator TODO
+		Propagators propagation.Propagators
 	}
 }
 
--- a/v4/newrelic/config.go
+++ b/v4/newrelic/config.go
@@ -341,7 +341,7 @@ type Config struct {
 	OpenTelemetry struct {
 		// Tracer TODO
 		Tracer trace.Tracer
-		// Propagator TODO
+		// Propagators TODO
 		Propagators propagation.Propagators
 	}
 }
--- a/v4/newrelic/instrumentation.go
+++ b/v4/newrelic/instrumentation.go
@@ -80,5 +80,37 @@ func WrapHandleFunc(app *Application, pattern string, handler func(http.Response
 // http.RoundTripper will look for a Transaction in the request's context
 // (using FromContext).
 func NewRoundTripper(original http.RoundTripper) http.RoundTripper {
-	return original
+	if nil == original {
+		original = http.DefaultTransport
+	}
+	return roundTripperFunc(func(request *http.Request) (*http.Response, error) {
+		// The specification of http.RoundTripper requires that the request is never modified.
+		request = cloneRequest(request)
+		segment := StartExternalSegment(nil, request)
+
+		response, err := original.RoundTrip(request)
+
+		segment.Response = response
+		segment.End()
+
+		return response, err
+	})
 }
+
+// cloneRequest mimics implementation of
+// https://godoc.org/github.com/google/go-github/github#BasicAuthTransport.RoundTrip
+func cloneRequest(r *http.Request) *http.Request {
+	// shallow copy of the struct
+	r2 := new(http.Request)
+	*r2 = *r
+	// deep copy of the Header
+	r2.Header = make(http.Header, len(r.Header))
+	for k, s := range r.Header {
+		r2.Header[k] = append([]string(nil), s...)
+	}
+	return r2
+}
+
+type roundTripperFunc func(*http.Request) (*http.Response, error)
+
+func (f roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) { return f(r) }
--- a/v4/newrelic/segments.go
+++ b/v4/newrelic/segments.go
@@ -331,7 +331,9 @@ func StartExternalSegment(txn *Transaction, request *http.Request) *ExternalSegm
 		Request:   request,
 	}
 
-	txn.InsertDistributedTraceHeaders(request.Header)
+	if nil != request && nil != request.Header {
+		txn.InsertDistributedTraceHeaders(request.Header)
+	}
 
 	return s
 }
--- a/v4/newrelic/segments.go
+++ b/v4/newrelic/segments.go
@@ -323,5 +323,15 @@ func (s *ExternalSegment) SetStatusCode(code int) {}
 // NewRoundTripper: You may not need to use StartExternalSegment at all!
 //
 func StartExternalSegment(txn *Transaction, request *http.Request) *ExternalSegment {
-	return nil
+	if nil == txn && nil != request {
+		txn = FromContext(request.Context())
+	}
+	s := &ExternalSegment{
+		StartTime: txn.StartSegmentNow(),
+		Request:   request,
+	}
+
+	txn.InsertDistributedTraceHeaders(request.Header)
+
+	return s
 }
--- a/v4/newrelic/segments_test.go
+++ b/v4/newrelic/segments_test.go
@@ -685,3 +685,30 @@ func TestStartExternalSegment(t *testing.T) {
 		t.Errorf("expected traceparent '%s', got '%s'", expectedTraceparent, traceparent)
 	}
 }
+
+func TestStartExternalSegmentWithTxnContext(t *testing.T) {
+	app := newTestApp(t)
+	txn := app.StartTransaction("transaction")
+	ctx := NewContext(context.Background(), txn)
+
+	req, _ := http.NewRequest("GET", "http://request.com/", nil)
+	req = req.WithContext(ctx)
+	seg1 := StartExternalSegment(nil, req)
+	seg1.End()
+
+	txn.End()
+
+	// This will currently fail, as seg1.StartTime is nil. It can be
+	// activated context support is implemented.
+	if false {
+		traceID := getTraceID(txn.rootSpan.Span)
+		seg1ID := getSpanID(seg1.StartTime.Span)
+
+		traceparent := req.Header.Get("traceparent")
+		expectedTraceparent := fmt.Sprintf("00-%s-%s-00", traceID, seg1ID)
+
+		if traceparent != expectedTraceparent {
+			t.Errorf("expected traceparent '%s', got '%s'", expectedTraceparent, traceparent)
+		}
+	}
+}
--- a/v4/newrelic/segments_test.go
+++ b/v4/newrelic/segments_test.go
@@ -5,6 +5,7 @@ package newrelic
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"testing"
 
@@ -661,5 +662,26 @@ func TestSpanKind(t *testing.T) {
 	}
 	if kind := getSpanKind(segMessage.StartTime.Span); kind != trace.SpanKindInternal {
 		t.Errorf("segMessage has incorrect SpanKind: %s", kind)
+	}
+}
+
+func TestStartExternalSegment(t *testing.T) {
+	app := newTestApp(t)
+	txn := app.StartTransaction("transaction")
+
+	req, _ := http.NewRequest("GET", "http://request.com/", nil)
+	seg1 := StartExternalSegment(txn, req)
+	seg1.End()
+
+	txn.End()
+
+	traceID := getTraceID(txn.rootSpan.Span)
+	seg1ID := getSpanID(seg1.StartTime.Span)
+
+	traceparent := req.Header.Get("traceparent")
+	expectedTraceparent := fmt.Sprintf("00-%s-%s-00", traceID, seg1ID)
+
+	if traceparent != expectedTraceparent {
+		t.Errorf("expected traceparent '%s', got '%s'", expectedTraceparent, traceparent)
 	}
 }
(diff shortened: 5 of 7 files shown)


v4/newrelic/transaction.go
```diff
@@ -214,7 +222,17 @@ func (txn *Transaction) InsertDistributedTraceHeaders(hdrs http.Header) {}
...
+func (txn *Transaction) AcceptDistributedTraceHeaders(t TransportType, hdrs http.Header) {
+	props := propagation.New(propagation.WithExtractors(trace.TraceContext{}))
+
+	// Here we create a OpenTelemetry context that is detached from the
+	// current trace. All segments (spans) subsequently started with this
+	// context will be detached from the transaction trace, but rather will
+	// have the remote trace id as trace id and the remote span id as the
+	// parent span id.
+	remoteCtx := propagation.ExtractHTTP(context.Background(), props, hdrs)
+	txn.thread.getCurrentSpan().ctx = remoteCtx
```
purple4reina, line 234: Does this need to happen under lock?

v4/newrelic/transaction.go
```diff
@@ -198,7 +202,11 @@ func (txn *Transaction) StartSegment(name string) *Segment {
 //
 // StartExternalSegment calls InsertDistributedTraceHeaders, so you don't need
 // to use it for outbound HTTP calls: Just use StartExternalSegment!
-func (txn *Transaction) InsertDistributedTraceHeaders(hdrs http.Header) {}
+func (txn *Transaction) InsertDistributedTraceHeaders(hdrs http.Header) {
+	props := propagation.New(propagation.WithInjectors(trace.TraceContext{}))
```
purple4reina, line 206: Is this something we can cache on the txn itself?

purple4reina: Or even the app???

v4/newrelic/transaction.go
purple4reina: Also, what about the root span? Should we use the remoteCtx there as well?

v4/newrelic/transaction.go
```diff
@@ -214,7 +222,17 @@ func (txn *Transaction) InsertDistributedTraceHeaders(hdrs http.Header) {}
 // AcceptDistributedTraceHeaders first looks for the presence of W3C trace
 // context headers.  Only when those are not found will it look for the New
 // Relic distributed tracing header.
-func (txn *Transaction) AcceptDistributedTraceHeaders(t TransportType, hdrs http.Header) {}
+func (txn *Transaction) AcceptDistributedTraceHeaders(t TransportType, hdrs http.Header) {
+	props := propagation.New(propagation.WithExtractors(trace.TraceContext{}))
+
+	// Here we create a OpenTelemetry context that is detached from the
```
purple4reina, line 228:
```suggestion
	// Here we create an OpenTelemetry context that is detached from the
```

v4/newrelic/transaction.go
```diff
@@ -214,7 +222,17 @@ func (txn *Transaction) InsertDistributedTraceHeaders(hdrs http.Header) {}
...
+	// current trace. All segments (spans) subsequently started with this
+	// context will be detached from the transaction trace, but rather will
```
purple4reina, line 230:
```suggestion
	// context will be detached from the distributed trace, but rather will
```

purple4reina:
At some point we also need to implement and test `StartExternalSegment` and `NewRoundTripper`

v4/newrelic/transaction.go
pyohannes: Good point, I put it on the app.

v4/newrelic/application.go
```diff
@@ -28,7 +30,10 @@ func (app *Application) StartTransaction(name string) *Transaction {
 		rootSpan: s,
 		thread: &thread{
 			currentSpan: s,
+			spanCount:   1,
 		},
+		app:  app,
```
purple4reina, line 35: Should this be a pointer? I'm not sure, so could you explain your choice here so I can learn something?

v4/newrelic/instrumentation.go
```diff
@@ -80,5 +80,37 @@ func WrapHandleFunc(app *Application, pattern string, handler func(http.Response
...
+		request = cloneRequest(request)
+		segment := StartExternalSegment(nil, request)
+
+		response, err := original.RoundTrip(request)
+
+		segment.Response = response
+		segment.End()
+
+		return response, err
+	})
```
purple4reina, line 97: I should never have deleted this code in the first place. Sorry about that.

v4/newrelic/transaction.go
```diff
@@ -214,11 +230,40 @@ func (txn *Transaction) InsertDistributedTraceHeaders(hdrs http.Header) {}
...
+	// root segment that has the proper remote parent and trace id.
+	remoteCtx := propagation.ExtractHTTP(context.Background(), txn.app.propagators, hdrs)
+
+	if txn.thread.spanCount == 1 {
+		ctx, sp := txn.app.tracer.Start(remoteCtx, txn.name)
+		txn.rootSpan = &span{
+			Span: sp,
+			ctx:  ctx,
+		}
+		txn.thread.currentSpan = txn.rootSpan
```
purple4reina, line 258: All this business should probably be done under lock I think?

v4/newrelic/transaction.go
```diff
@@ -214,11 +230,40 @@ func (txn *Transaction) InsertDistributedTraceHeaders(hdrs http.Header) {}
...
+	} else {
+		txn.thread.currentSpan.ctx = remoteCtx
```
purple4reina, line 260: I think all the swapping of currentSpan and rootSpan here should be done under lock.

pyohannes: It is, see [line 238](https://github.com/newrelic/go-agent/pull/191/files#diff-6b77bc7503ec9881e453708a4bb0f3b1R238). I used the mutex on `txn.thread.currentSpan`. I'm not sure if this merits introducing another mutex on the transaction itself.

purple4reina: Ohhh duh! I'm blind. Thanks!

v4/newrelic/transaction.go
```diff
@@ -214,11 +230,40 @@ func (txn *Transaction) InsertDistributedTraceHeaders(hdrs http.Header) {}
...
+	txn.thread.currentSpan.Lock()
+	defer txn.thread.currentSpan.Unlock()
+
+	// Here we create a OpenTelemetry context that is detached from the
+	// current trace. All segments (spans) subsequently started with this
+	// context will be detached from the transaction trace, but rather will
+	// have the remote trace id as trace id and the remote span id as the
+	// parent span id.
+	//
+	// If no more than the root segment where yet created for this
```
purple4reina, line 247:
```suggestion
	// If no more than the root segment were yet created for this
```

v4/newrelic/segments_test.go
```diff
@@ -374,3 +376,24 @@ func TestSegmentsOnEmptyTransaction(t *testing.T) {
 	seg.End()
 	txn.End()
 }
+
+func TestStartExternalSegment(t *testing.T) {
+	app := newTestApp(t)
+	txn := app.StartTransaction("transaction")
+
+	req, _ := http.NewRequest("GET", "http://request.com/", nil)
+	seg1 := StartExternalSegment(txn, req)
```
purple4reina, line 385: We should probably also include a test that has the txn in the req's context. However, that test will currently fail because we haven't yet implemented `FromContext`.

pyohannes: I added a test, the test conditions being commented out (as they currently fail). It can be activated once context support is merged.

v4/newrelic/transaction.go
```diff
@@ -18,14 +20,17 @@ import (
...
 	rootSpan *span
 	thread   *thread
 	ended    bool
+	name     string
 }
 
 type thread struct {
 	sync.Mutex
 	currentSpan *span
+	spanCount   uint
```
purple4reina, line 33: We could probably make this a bool with the same effect.

v4/newrelic/transaction.go
```diff
@@ -260,6 +304,7 @@ func (txn *Transaction) NewGoroutine() *Transaction {
 	newTxn := *txn
 	newTxn.thread = &thread{
 		currentSpan: txn.thread.currentSpan,
+		spanCount:   1,
```
purple4reina, line 307: I feel like this number should match that on the `txn.thread`.

purple4reina: Here's an example app that I think demonstrates my thinking:

```go
package main

import (
	"log"
	"net/http"
	"time"

	"github.com/newrelic/go-agent/v4/newrelic"
	"go.opentelemetry.io/otel/exporters/stdout"
)

func main() {
	pusher, err := stdout.InstallNewPipeline(nil, nil)
	if err != nil {
		log.Fatal(err)
	}
	defer pusher.Stop()

	app, _ := newrelic.NewApplication(
		newrelic.ConfigFromEnvironment(),
	)
	txn := app.StartTransaction("main")
	defer txn.End()

	txn.StartSegment("AAAAAAAAAAAAAAAAAA").End()

	go func(txn *newrelic.Transaction) {
		hdr := http.Header{}
		hdr.Set("traceparent", "00-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-bbbbbbbbbbbbbbbb-01")
		txn.AcceptDistributedTraceHeaders("HTTP", hdr)
		txn.StartSegment("CCCCCCCCCCCCCCCCCC").End()
	}(txn.NewGoroutine())

	time.Sleep(time.Second)

	txn.StartSegment("BBBBBBBBBBBBBBBBBB").End()
}
```

As currently implemented, the Accept call will swap out the root span and therefore segment C will have a different trace ID than A and B.

v4/newrelic/transaction.go
```diff
@@ -260,6 +304,7 @@ func (txn *Transaction) NewGoroutine() *Transaction {
 	newTxn := *txn
```
purple4reina, line 304: This doesn't need to be done under lock does it?

v4/newrelic/segments.go
```diff
@@ -323,5 +323,15 @@ func (s *ExternalSegment) SetStatusCode(code int) {}
...
-	return nil
+	if nil == txn && nil != request {
+		txn = FromContext(request.Context())
+	}
+	s := &ExternalSegment{
+		StartTime: txn.StartSegmentNow(),
+		Request:   request,
+	}
+
+	txn.InsertDistributedTraceHeaders(request.Header)
```
purple4reina, line 334: I think we need to check for a nil request and nil headers here. That's what we do in the v3 agent at least.

purple4reina: Or at least we should probably have tests for those two cases.

v4/newrelic/transaction.go
```diff
@@ -204,7 +210,16 @@ func (txn *Transaction) StartSegment(name string) *Segment {
...
-func (txn *Transaction) InsertDistributedTraceHeaders(hdrs http.Header) {}
+func (txn *Transaction) InsertDistributedTraceHeaders(hdrs http.Header) {
+	if txn == nil {
+		return
+	}
+
+	txn.thread.currentSpan.Lock()
+	defer txn.thread.currentSpan.Unlock()
+
+	propagation.InjectHTTP(txn.thread.currentSpan.ctx, txn.app.propagators, hdrs)
```
purple4reina, line 221: I think I have a method already called `getCurrentSpan` that does the locking and everything 💜 🟣 🟪

purple4reina: Also, I believe this line will panic in this case:

```go
txn := &Transaction{}
txn.InsertDistributedTraceHeaders(hdrs)
```

v4/newrelic/transaction.go
```diff
@@ -220,11 +235,40 @@ func (txn *Transaction) InsertDistributedTraceHeaders(hdrs http.Header) {}
...
 // context headers.  Only when those are not found will it look for the New
 // Relic distributed tracing header.
-func (txn *Transaction) AcceptDistributedTraceHeaders(t TransportType, hdrs http.Header) {}
+func (txn *Transaction) AcceptDistributedTraceHeaders(t TransportType, hdrs http.Header) {
+	if txn == nil {
+		return
+	}
+
+	txn.thread.currentSpan.Lock()
+	defer txn.thread.currentSpan.Unlock()
```
purple4reina, line 244: Similarly, this line will panic if the transaction is a `&Transaction{}`.

v4/newrelic/transaction.go
```diff
@@ -220,11 +235,40 @@ func (txn *Transaction) InsertDistributedTraceHeaders(hdrs http.Header) {}
...
+		txn.thread.currentSpan = txn.rootSpan
+	} else {
+		txn.thread.currentSpan.ctx = remoteCtx
+	}
+}
 
 // Application returns the Application which started the transaction.
 func (txn *Transaction) Application() *Application {
-	return nil
+	return txn.app
```
purple4reina, line 271: :hooray: so easy!

v4/newrelic/transaction.go
pyohannes: Fixed, and added a test.

v4/newrelic/transaction.go
pyohannes: True. Fixed and added a test.

v4/newrelic/segments.go
pyohannes: Agreed. Fixed and added a test.

v4/newrelic/transaction.go
pyohannes: You're right, it should be under a lock. We're copying information from `txn` in a non-atomic way.

v4/newrelic/transaction.go
pyohannes: You're right. I fixed this and added a test.

v4/newrelic/transaction.go
pyohannes: 👍

v4/newrelic/application.go
```diff
@@ -104,5 +109,9 @@ func NewApplication(opts ...ConfigOption) (*Application, error) {
 	if nil == tracer {
 		tracer = global.Tracer("traceName")
 	}
-	return &Application{tracer: tracer}, nil
+	propagators := propagation.New(
+		propagation.WithInjectors(trace.TraceContext{}),
+		propagation.WithExtractors(trace.TraceContext{}))
```
purple4reina, line 114: Just thinking aloud here. In the future it might be nice to allow customers to configure which injector/extractor they use. We could put the `trace.TraceContext` as a field on the `Config.OpenTelemetry` config option. Something like that.

I say this because a while ago I implemented a B3 propagation integration for the agent. In v4 of the agent, we could deprecate that integration and instead allow customers to specifically set their injector/extractor.

v4/newrelic/transaction.go
```diff
@@ -257,9 +304,13 @@ func (txn *Transaction) BrowserTimingHeader() *BrowserTimingHeader {
...
 // be reported.
 func (txn *Transaction) NewGoroutine() *Transaction {
+	txn.thread.Lock()
+	defer txn.thread.Unlock()
+
 	newTxn := *txn
 	newTxn.thread = &thread{
-		currentSpan: txn.thread.currentSpan,
+		currentSpan:  txn.thread.currentSpan,
+		isSingleSpan: txn.thread.isSingleSpan,
```
purple4reina, line 313: Just thinking aloud. I think there's more we need to consider.

```go
package main

import (
	"log"
	"net/http"
	"time"

	"github.com/newrelic/go-agent/v4/newrelic"
	"go.opentelemetry.io/otel/exporters/stdout"
)

func main() {
	pusher, err := stdout.InstallNewPipeline(nil, nil)
	if err != nil {
		log.Fatal(err)
	}
	defer pusher.Stop()

	app, _ := newrelic.NewApplication(
		newrelic.ConfigFromEnvironment(),
	)
	txn := app.StartTransaction("main")
	defer txn.End()

	segFunc := func(name string, txn *newrelic.Transaction) {
		defer txn.StartSegment(name).End()
	}

	go segFunc("seg1", txn.NewGoroutine())
	time.Sleep(time.Second)

	go func(txn *newrelic.Transaction) {
		hdr := http.Header{}
		hdr.Set("traceparent", "00-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-bbbbbbbbbbbbbbbb-01")
		txn.AcceptDistributedTraceHeaders("HTTP", hdr)
		segFunc("seg2", txn)
	}(txn.NewGoroutine())
	time.Sleep(time.Second)

	segFunc("seg3", txn)
}
```

The first segment and the third segment will be reported, but the second one will be orphaned. This is because the new root span that is created when Accept is called is never ended. When `txn.End()` is called, it is ending and finalizing the original root span.

Possible solution: disallow swapping of root spans if Accept is called inside a goroutine (aka `txn.NewGoroutine` was the creator of the `txn.thread`).

Thoughts?

v4/newrelic/application.go
```diff
@@ -28,8 +30,11 @@ func (app *Application) StartTransaction(name string) *Transaction {
 	return &Transaction{
 		rootSpan: s,
 		thread: &thread{
-			currentSpan: s,
+			currentSpan:  s,
+			isSingleSpan: true,
```
purple4reina, line 34: Nitpick. If instead we use a bool that starts as `false` and moves to `true` when a span is started, we can rely on the default value of the bool. This might be helpful if we were to accidentally not include this field in the creation of a new `*thread`.

v4/newrelic/transaction.go
```diff
@@ -18,14 +20,17 @@ import (
...
 	thread   *thread
 	ended    bool
+	name     string
 }
 
 type thread struct {
 	sync.Mutex
-	currentSpan *span
+	currentSpan  *span
+	isSingleSpan bool
```
purple4reina, line 33: Nitpick. Since the usage of this field isn't completely self evident, could we add a comment above it?

v4/newrelic/transaction.go
pyohannes: Good point. Is it a common practice (or even allowed) to call `txn.End` on the transaction that was returned from `NewGoroutine`? Or do we anticipate that this call always happens on the original transaction?

v4/newrelic/application.go
pyohannes: Done. It is now `isMultiSpan`.

v4/newrelic/transaction.go
pyohannes: 👍

v4/newrelic/application.go
pyohannes: Done. I put the propagators into the config, along with the tracer. This should give users full flexibility. I also added a B3 test.

v4/newrelic/transaction.go
purple4reina: We have always touted our `*Transaction` as being thread safe and thus that you can call `txn.End` anywhere :(

v4/newrelic/config.go
```diff
@@ -340,6 +341,8 @@ type Config struct {
 	OpenTelemetry struct {
 		// Tracer TODO
 		Tracer trace.Tracer
+		// Propagator TODO
```
purple4reina, line 344: nitpick:

```suggestion
		// Propagators TODO
```

v4/newrelic/transaction.go
pyohannes: > Possible solution: disallow swapping of root spans if Accept is called inside a goroutine (aka txn.NewGoroutine was the creator of the txn.thread).

I think to be safe we have to go a little bit further and disallow root span swapping in general once `NewGoroutine` was called, for the original transaction as well as for the one created via `NewGoroutine`. We have to prevent that transactions have different root segments, otherwise calls to `txn.End` might cause orphaned spans.

pyohannes: I pushed a commit implementing that.

v4/newrelic/transaction.go
```diff
@@ -257,9 +306,15 @@ func (txn *Transaction) BrowserTimingHeader() *BrowserTimingHeader {
 // Note that any segments that end after the transaction ends will not
 // be reported.
 func (txn *Transaction) NewGoroutine() *Transaction {
+	txn.thread.Lock()
+	defer txn.thread.Unlock()
+
+	txn.thread.isMultiSpan = true
```
purple4reina, line 312: Oh nice, I hadn't thought of this change but it makes total sense. This is great.

v4/newrelic/transaction.go
```diff
@@ -18,14 +20,19 @@ import (
...
 	thread   *thread
 	ended    bool
+	name     string
 }
 
 type thread struct {
 	sync.Mutex
 	currentSpan *span
+	// isMultiSpan is set to true of the associated transaction has more
+	// than one span.
```
purple4reina, line 34:
```suggestion
	// than one span or crosses multiple goroutines.
```

PR #192, opened 2020-07-29
Set SpanKind to internal for all spans.

purple4reina:
With every span we need to also include the [Span Kind](https://pkg.go.dev/go.opentelemetry.io/otel@v0.9.0/api/trace?tab=doc#SpanKind). Unfortunately right now, OTel's Go implementation only supports setting the `SpanKind` at span start time. Since we cannot know which type of segment `txn.StartSegmentNow` is being called for, this will not work for us. This PR sets all the spans we create to be "internal" spans.

I spoke with @MrAlias about this and he opened an issue https://github.com/open-telemetry/opentelemetry-go/issues/988 that would mean we could set the `SpanKind` at either start _or_ end. Once that work is done, we need to circle back and implement the span kind for each span. We'll need to be sure to set the root span to the "server" type iff the `txn.SetWebRequest` API has been called, otherwise it should be left as "internal".

Lastly, this PR is reliant on https://github.com/open-telemetry/opentelemetry-go/pull/987 being merged into OpenTelemetry. The tests will fail until that happens. In the meantime, it is still ready for your review.

### DIFF
--- a/build-script.sh
+++ b/build-script.sh
@@ -36,7 +36,7 @@ for dir in $DIRS; do
   if [ $dir == "." ]; then
     rm -rf v3/
     rm -rf v4/
-  elif [ $dir = v3* ]; then
+  elif [[ $dir =~ v3 ]]; then
     # Only v3 code version 1.9+ needs GRPC dependencies
     VERSION=$(go version)
     V17="1.7"
@@ -50,7 +50,7 @@ for dir in $DIRS; do
     fi
   else
     # install v4 dependencies
-    go get ./...
+    go get -u go.opentelemetry.io/otel
   fi
 
   go test -race -benchtime=1ms -bench=. ./...
--- a/build-script.sh
+++ b/build-script.sh
@@ -50,7 +50,7 @@ for dir in $DIRS; do
     fi
   else
     # install v4 dependencies
-    go get -u go.opentelemetry.io/otel
+    go get ./...
   fi
 
   go test -race -benchtime=1ms -bench=. ./...
--- a/v4/newrelic/segments_test.go
+++ b/v4/newrelic/segments_test.go
@@ -28,9 +28,14 @@ func getSpanName(s trace.Span) string {
 	return s.(*testtrace.Span).Name()
 }
 
+func getSpanKind(s trace.Span) trace.SpanKind {
+	return s.(*testtrace.Span).SpanKind()
+}
+
 func newTestApp(t *testing.T) *Application {
 	app, err := NewApplication(func(cfg *Config) {
-		cfg.OpenTelemetry.Tracer = testtrace.NewTracer()
+		tp := testtrace.NewProvider()
+		cfg.OpenTelemetry.Tracer = tp.Tracer("go-agent-test")
 	})
 	if err != nil {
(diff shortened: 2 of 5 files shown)


build-script.sh
```diff
@@ -50,7 +50,7 @@ for dir in $DIRS; do
     fi
   else
     # install v4 dependencies
-    go get ./...
+    go get -u go.opentelemetry.io/otel
```
purple4reina, line 53: My PR to opentelemetry is included in v0.10.0, however doing a `go get ./...` installs what's currently listed in the go.mod file, which is v0.5.0.

v4/newrelic/segments_test.go
```diff
@@ -23,9 +24,18 @@ func spanHasEnded(s trace.Span) bool {
...
+
+func getSpanKind(s trace.Span) trace.SpanKind {
+	return s.(*testtrace.Span).SpanKind()
+}
+
 func newTestApp(t *testing.T) *Application {
 	app, err := NewApplication(func(cfg *Config) {
-		cfg.OpenTelemetry.Tracer = testtrace.NewTracer()
+		tp := testtrace.NewProvider()
+		cfg.OpenTelemetry.Tracer = tp.Tracer("go-agent-test")
```
purple4reina, line 38: This change was required by the upgrade to the latest opentelemetry version.

build-script.sh
```diff
@@ -36,7 +36,7 @@ for dir in $DIRS; do
   if [ $dir == "." ]; then
     rm -rf v3/
     rm -rf v4/
-  elif [ $dir = v3* ]; then
+  elif [[ $dir =~ v3 ]]; then
```
purple4reina, line 39: Sadly, this check wasn't correctly matching. It works now.

build-script.sh
pyohannes: Why don't we update the `go.mod` file to `0.10.0`?

pyohannes: With the change in the `go.mod` file, do we still need the `go get -u go.opentelemetry.io/otel` line?

purple4reina: Yeah, I was thinking about that.

With `go get -u go.opentelemetry.io/otel` we will always test against the latest version of otel.

With `go get ./...` we will test against whatever is written in the go mod file (I think, this might depend on go version, so I'm not 100% certain).

Personally, I'd rather test against the latest version of the otel package so we can catch changes as quickly as we can. However, then we could end up being out of sync with our go.mod file.

So basically, I'm not sure what the right answer is. What do you think?

pyohannes: I'll try to imagine the worst possible scenario:
1. Our `go.mod` file references an older version `0.x`.
2. Our tests pull the newest version `0.y`.
3. We modify our code to work with `0.y` but don't update the `go.mod` file. Unfortunately, those are breaking changes, so the agent won't work anymore with `0.x`.
4. We release.
5. It breaks for customers.

I'm not 100% familiar with all the `go.mod` delicacies, but I'd opt for running the tests with the same version we have in `go.mod`.

purple4reina: Very good point. I'll make the update!

PR #195, opened 2020-08-03
Support testing integration packages in v4.

purple4reina:
This Pull Request enables most of the tests on our integration packages. The ones that are not enabled are given clear comments as to why they are not enabled and what it will take to enable them.

At this point, any call to any of our `Expect*` methods (ex `ExpectSpanEvents`) does nothing. This means that mostly these tests are just making sure packages compile. I will implement two of these expect methods in my next Pull Request.

I have commented throughout the code for easier reviewing. 💜


v4/integrations/nrawssdk-v1/nrawssdk_test.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 func testApp() integrationsupport.ExpectApp {
-	return integrationsupport.NewTestApp(integrationsupport.SampleEverythingReplyFn, integrationsupport.DTEnabledCfgFn)
+	return integrationsupport.NewTestApp(integrationsupport.DTEnabledCfgFn)
```
purple4reina, line 28: I changed the signature on `NewTestApp` because the first argument was a `func(reply *internal.ConnectReply)`. Since we no longer have anything that has anything to do with connect, this argument is now useless. So I removed it.

v4/internal/awssupport/awssupport.go
```diff
@@ -0,0 +1,105 @@
+// Copyright 2020 New Relic Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// +build go1.8
+
+package awssupport
```
purple4reina, line 6: This package is required for the two sets of `aws-sdk` tests. All I did was copy the package from the v3 directory and change all instances of `v3` to `v4`.

v4/internal/integrationsupport/integrationsupport.go
```diff
@@ -0,0 +1,73 @@
...
+
+	app, err := newrelic.NewApplication(cfgFn...)
+	if nil != err {
+		panic(err)
+	}
+
+	return ExpectApp{
+		Expect: &internal.TraceExpecter{
+			Tracer: tr.(*testtrace.Tracer),
+		},
```
purple4reina, line 54: This is the most important and complex change in this PR. Ultimately, we want our current integrations tests to still be able to run with very minimal code changes. These tests all involve calling `integrationsupport.NewTestApp` then calling the expect methods on the returned app. For example `app.ExpectSpanEvents` or `app.ExpectMetrics`.

In the v3 incarnation of the agent, the implementation of these `app.Expect*` methods were in the `newrelic` package proper and were compiled into agent binaries. (Maybe they were compiled out by the compiler, but either way, they lived in the agent API code proper) I've always felt this was a mistake. So, instead I've chosen to put that code in `internal` instead as part of a new `internal.TraceExpecter` type. This way it will only ever be used when we are specifically testing.

v4/internal/expect.go
```diff
@@ -0,0 +1,111 @@
+// Copyright 2020 New Relic Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package internal
```
purple4reina, line 4: This file was copied from the `v3/internal/expect.go` file. I removed a couple of methods that had to do with connect replies and harvests. Otherwise, I changed nothing.

v4/internal/sysinfo/hostname.go
```diff
@@ -0,0 +1,9 @@
+// Copyright 2020 New Relic Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package sysinfo
```
purple4reina, line 4: The `sysinfo` package is required for testing `nrmongo`.

v4/internal/trace_expecter.go
```diff
@@ -0,0 +1,40 @@
+package internal
```
purple4reina, line 1: This file is where the birthday cake ice cream lies.

The `OpenTelemetryExpect` in this file wraps a `*testtrace.Tracer` and adds all the methods required for `internal.Expect`. This way we can easily access the data saved to the Tracer when testing. In the future when we add otel metrics to the shim, we can simply imbed the otel metric type here as well.

v4/internal/trace_expecter.go
```diff
@@ -0,0 +1,40 @@
...
+// OpenTelemetryExpect implements internal.Expect for use in testing.
+type OpenTelemetryExpect struct {
+	*testtrace.Tracer
+}
+
+// ExpectTxnEvents TODO
+func (e *OpenTelemetryExpect) ExpectTxnEvents(t Validator, want []WantEvent) {}
+
+// ExpectSpanEvents TODO
+func (e *OpenTelemetryExpect) ExpectSpanEvents(t Validator, want []WantEvent) {}
```
purple4reina, line 16: Currently all the expect methods do nothing. I chose to do this work in stages. The next Pull Request I plan on opening will implement `ExpectTxnEvents` and `ExpectSpanEvents`.

v4/newrelic/transaction.go
```diff
@@ -126,6 +132,15 @@ func (txn *Transaction) SetWebRequest(r WebRequest) {}
...
 func (txn *Transaction) SetWebResponse(w http.ResponseWriter) http.ResponseWriter {
+	if w == nil {
+		// Accepting a nil parameter makes it easy for consumers to add
+		// a response code to the transaction without a response
+		// writer:
+		//
+		//    txn.SetWebResponse(nil).WriteHeader(500)
+		//
+		w = dummyResponseWriter{}
+	}
```
purple4reina, line 143: I believe it's the `nrecho` integration that uses `txn.SetWebResponse(nil).WriteHeader(code)`. Those tests fail without this addition.

v4/internal/otel_expect.go
```diff
@@ -0,0 +1,40 @@
...
+func (e *OpenTelemetryExpect) ExpectMetricsPresent(t Validator, want []WantMetric) {}
+
+// ExpectTxnMetrics TODO
+func (e *OpenTelemetryExpect) ExpectTxnMetrics(t Validator, want WantTxn) {}
+
+// ExpectTxnTraces TODO
+func (e *OpenTelemetryExpect) ExpectTxnTraces(t Validator, want []WantTxnTrace) {}
+
+// ExpectSlowQueries TODO
+func (e *OpenTelemetryExpect) ExpectSlowQueries(t Validator, want []WantSlowQuery) {}
```
pyohannes, line 40: Thanks for providing this. Just to make sure I understood correctly: I assume from these we will only implement `ExpectSpanEvents`, `ExpectMetrics`, `ExpectMetricsPresent` and `ExpectTxnMetrics`?

purple4reina: I'm glad you brought that up, cuz I was thinking of implementing different ones than that! I had written up comments that included that information, but it seems the disappeared when I later made some changes. I'll re-add them.

v4/internal/otel_expect.go
```diff
@@ -0,0 +1,40 @@
...
+// OpenTelemetryExpect implements internal.Expect for use in testing.
+type OpenTelemetryExpect struct {
+	*testtrace.Tracer
+}
+
+// ExpectTxnEvents TODO
+func (e *OpenTelemetryExpect) ExpectTxnEvents(t Validator, want []WantEvent) {}
+
+// ExpectSpanEvents TODO
+func (e *OpenTelemetryExpect) ExpectSpanEvents(t Validator, want []WantEvent) {}
```
purple4reina, line 16: Currently all the expect methods do nothing. I chose to do this work in stages. The next Pull Request I plan on opening will implement `ExpectTxnEvents` and `ExpectSpanEvents`. I can see we may consider implementing `ExpectTxnMetrics` as well, but since we aren't actually going to be creating any metrics, I'm a bit iffy on it.

v4/internal/otel_expect.go
```diff
@@ -0,0 +1,40 @@
+package internal
```
purple4reina, line 1: This file is where the birthday cake ice cream lies.

The `OpenTelemetryExpect` in this file wraps a `*testtrace.Tracer` and adds all the methods required for `internal.Expect`. This way we can easily access the data saved to the Tracer when testing. In the future when we add otel metrics to the shim, we can simply imbed the otel metric type here as well.

.github/workflows/ci.yaml
```diff
@@ -213,6 +213,120 @@ jobs:
...
+            dirs: v4/integrations/nrredis-v7
+            extratesting: go get -u github.com/go-redis/redis/v7@master
+          - go-version: 1.14.x
+            dirs: v4/integrations/nrsqlite3
+            extratesting: go get -u github.com/mattn/go-sqlite3@master
+          - go-version: 1.14.x
+            dirs: v4/integrations/nrsnowflake
+            extratesting: go get -u github.com/snowflakedb/gosnowflake@master
+          # gRPC tests should be skipped until Distributed Tracing has been
+          # implemented.
```
pyohannes, line 279: Could we activate all those tests now?

purple4reina: Yup! I'd rather do it in a separate PR since it will take a bit of work.

PR #196, opened 2020-08-04
Implement ExpectSpanEvents.

purple4reina:
This pull request implements `ExpectSpanEvents` for use when testing integration packages. It also removes `ExpectTxnEvents` because those are essentially just span events anyway, so the APIs have been combined.

I'd love your feedback on the use of all the `internal.Match*` fields. I'd like your thoughts on what the best user experience of these would be.

### DIFF
--- a/v4/internal/expect.go
+++ b/v4/internal/expect.go
@@ -6,6 +6,8 @@ package internal
 // Validator is used for testing.
 type Validator interface {
 	Error(...interface{})
+	Errorf(string, ...interface{})
+	Helper()
 }
 
 // WantMetric is a metric expectation.  If Data is nil, then any data values are
@@ -34,8 +36,12 @@ func uniquePointer() *struct{} {
 var (
 	// MatchAnything is for use when matching attributes.
 	MatchAnything = uniquePointer()
-	// MatchAnyString is a placeholder for matching any string
-	MatchAnyString = "xxANY-STRINGxx"
+	// MatchNoParent is for use when matching ParentID on spans.  It only
+	// matches for spans with no parent.
+	MatchNoParent = "0000000000000000"
+	// MatchAnyParent is for use when matching ParentID on spans.  It matches
+	// any ParentID value except for the no parent value.
+	MatchAnyParent = "💜💜xxANY-PARENTxx💜💜"
 )
 
 // WantEvent is a transaction or error event expectation.
@@ -91,21 +97,25 @@ type WantTxn struct {
 	NumErrors int
 }
 
+// WantSpan is a span or transaction expectation.
+type WantSpan struct {
+	Name       string
+	SpanID     string
+	TraceID    string
+	ParentID   string
+	Attributes map[string]interface{}
+}
+
 // Expect exposes methods that allow for testing whether the correct data was
 // captured.
 type Expect interface {
 	ExpectCustomEvents(t Validator, want []WantEvent)
 	ExpectErrors(t Validator, want []WantError)
 	ExpectErrorEvents(t Validator, want []WantEvent)
-
-	ExpectTxnEvents(t Validator, want []WantEvent)
-
 	ExpectMetrics(t Validator, want []WantMetric)
 	ExpectMetricsPresent(t Validator, want []WantMetric)
 	ExpectTxnMetrics(t Validator, want WantTxn)
-
 	ExpectTxnTraces(t Validator, want []WantTxnTrace)
 	ExpectSlowQueries(t Validator, want []WantSlowQuery)
-
-	ExpectSpanEvents(t Validator, want []WantEvent)
+	ExpectSpanEvents(t Validator, want []WantSpan)
 }
--- a/v4/internal/integrationsupport/integrationsupport.go
+++ b/v4/internal/integrationsupport/integrationsupport.go
@@ -38,8 +38,9 @@ func ConfigFullTraces(cfg *newrelic.Config) {
 
 // NewTestApp creates an ExpectApp with the given ConnectReply function and Config function
 func NewTestApp(cfgFn ...newrelic.ConfigOption) ExpectApp {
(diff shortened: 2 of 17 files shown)


v4/internal/expect.go
```diff
@@ -91,21 +99,25 @@ type WantTxn struct {
...
+
 // Expect exposes methods that allow for testing whether the correct data was
 // captured.
 type Expect interface {
 	ExpectCustomEvents(t Validator, want []WantEvent)
 	ExpectErrors(t Validator, want []WantError)
 	ExpectErrorEvents(t Validator, want []WantEvent)
-
-	ExpectTxnEvents(t Validator, want []WantEvent)
-
```
purple4reina, line 102: ExpectTxnEvents was removed in favor of ExpectSpanEvents.

v4/internal/expect.go
```diff
@@ -91,21 +99,25 @@ type WantTxn struct {
...
 	ExpectMetrics(t Validator, want []WantMetric)
 	ExpectMetricsPresent(t Validator, want []WantMetric)
 	ExpectTxnMetrics(t Validator, want WantTxn)
-
 	ExpectTxnTraces(t Validator, want []WantTxnTrace)
 	ExpectSlowQueries(t Validator, want []WantSlowQuery)
-
-	ExpectSpanEvents(t Validator, want []WantEvent)
+	ExpectSpanEvents(t Validator, want []WantSpan)
```
purple4reina, line 122: Instead of a WantEvent, this now accepts a WantSpan which has several new fields that WantEvent did not have.

v4/internal/integrationsupport/integrationsupport.go
```diff
@@ -50,7 +51,7 @@ func NewTestApp(cfgFn ...newrelic.ConfigOption) ExpectApp {
 
 	return ExpectApp{
 		Expect: &internal.OpenTelemetryExpect{
-			Tracer: tr.(*testtrace.Tracer),
+			Spans: sr,
```
purple4reina, line 54: This change is needed so we can capture the spans that have been recorded.

v4/internal/otel_expect.go
```diff
@@ -6,14 +6,66 @@ import (
 
 // OpenTelemetryExpect implements internal.Expect for use in testing.
 type OpenTelemetryExpect struct {
-	*testtrace.Tracer
+	Spans *testtrace.StandardSpanRecorder
 }
 
-// ExpectTxnEvents TODO
-func (e *OpenTelemetryExpect) ExpectTxnEvents(t Validator, want []WantEvent) {}
+func expectSpan(t Validator, want WantSpan, span *testtrace.Span) {
```
purple4reina, line 12: I'd love your feedback on the implementation of this method.

Basically, I made it so that each field on the `WantSpan` is required. If you specifically want a field to be ignored, you must specify that with the `internal.MatchAnyID` option. An alternative would be to ignore testing of a field if it is found to be an empty string on the `WantSpan`.

v4/internal/otel_expect.go
```diff
@@ -6,14 +6,66 @@ import (
...
+					noParent, id)
+			}
+		} else if want.ParentID == MatchAnyParent {
+			if id == noParent {
+				t.Error("Incorrect parent id: expected a parent but found none")
+			}
+		} else if id != want.ParentID {
+			t.Errorf("Incorrect parent id:\n\texpect=%s actual=%s",
+				want.ParentID, id)
+		}
```
purple4reina, line 48: For the `ParentID` field I implemented two additional options for the cases where the span has no parent and the case where the span has a parent but we don't care what the specific parent ID is. Thoughts here welcome as well.

v4/internal/expect.go
```diff
@@ -34,8 +36,14 @@ func uniquePointer() *struct{} {
...
-	// MatchAnyString is a placeholder for matching any string
-	MatchAnyString = "xxANY-STRINGxx"
+	// MatchAnyID is a placeholder for matching any Span or Trace ID.
+	MatchAnyID = "💜💜xxANY-IDxx💜💜"
+	// MatchNoParent is for use when matching ParentID on spans.  It only
+	// matches for spans with no parent.
+	MatchNoParent = "💜💜xxNO-PARENTxx💜💜"
+	// MatchAnyParent is for use when matching ParentID on spans.  It matches
+	// any ParentID value except for the no parent value.
+	MatchAnyParent = "💜💜xxANY-PARENTxx💜💜"
```
purple4reina, line 46: I used emoji so it would be very difficult to accidentally match these strings.

v4/internal/expect.go
```diff
@@ -91,21 +101,25 @@ type WantTxn struct {
 	NumErrors int
 }
 
+// WantSpan is a span or transaction expectation.
+type WantSpan struct {
+	Name       string
+	SpanID     string
```
pyohannes, line 107: Just an idea:

Could we use `interface{}` as type for those attributes? `nil` could mean "any" value. And we could have separate types for constants (e. g. like `MatchNoParent = MatchNoParentType{}`).

Using `nil` for "any" value would help to avoid explicit initialization of those values in most cases.

v4/internal/otel_expect.go
```diff
@@ -6,14 +6,62 @@ import (
...
+				name, want.ParentID, id)
+		}
+	}
+}
+
+func (e *OpenTelemetryExpect) spans() []*testtrace.Span {
+	return e.Spans.Completed()
+}
 
 // ExpectSpanEvents TODO
```
pyohannes, line 52: We can remove the TODO.

v4/internal/otel_expect.go
pyohannes: > An alternative would be to ignore testing of a field if it is found to be an empty string on the WantSpan.

I like this alternative better, thanks for switching to it. It seems more intuitive to me (we only test for what is explicitly specified).

PR #201, opened 2020-08-07
Modify Serverless to use a named pipe for telemetry when present

MattWhelan:
Serverless telemetry is normally sent to stdout. This is an alternate
path designed for use with an externally managed named pipe at a
fixed location.


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # ChangeLog
 
+## 3.8.2
```
purple4reina, line 3: In the past we have left the version out until right before we release. That way we can decide the release version later on.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
...
+
+### Changes
+* When sending Serverless telemetry, support an externally-managed named pipe.
```
purple4reina, line 6:
```suggestion
* When sending Serverless telemetry using the `nrlambda` integration, support an externally-managed named pipe.
```

v3/integrations/nrlambda/handler.go
```diff
@@ -67,14 +67,36 @@ func responseEvent(ctx context.Context, event interface{}) {
...
+type writerProvider interface {
+	borrowWriter(needsWriter func(writer io.Writer))
+}
+
+type defaultWriterProvider struct {
+}
+
+func (wp *defaultWriterProvider) borrowWriter(needsWriter func(io.Writer)) {
+	// If the telemetry named pipe exists and is writable, use it instead of stdout
+	const telemetryNamedPipe = "/tmp/newrelic-telemetry"
```
purple4reina, line 79: Should we make this a global const so it doesn't need to be redefined each time?

v3/integrations/nrlambda/handler.go
```diff
@@ -112,7 +134,7 @@ type wrappedHandler struct {
 	firstTransaction sync.Once
 	// writer is used to log the data JSON at the end of each transaction.
 	// This field exists (rather than hardcoded os.Stdout) for testing.
```
purple4reina, line 136: This comment needs to be updated.

v3/integrations/nrlambda/handler.go
MattWhelan: Doesn't the compiler do that for consts?

purple4reina: Oh, that I'm not sure about! Good question, I'll do some research.

purple4reina: Ah ha! You are correct! I learned something today! https://qvault.io/2019/10/14/constants-in-go-vs-javascript-and-when-to-use-them/

PR #202, opened 2020-08-07
Implement ExpectMetrics and ExpectMetricsPresent.

purple4reina:
This pull request implements two more Expect methods. This work also uncovered some bugs! 🎉 I fixed those in here as well. I will comment throughout the code to help with review.

### DIFF
--- a/v4/integrations/nrawssdk-v1/nrawssdk_test.go
+++ b/v4/integrations/nrawssdk-v1/nrawssdk_test.go
@@ -78,7 +78,7 @@ var (
 		},
 	}
 	externalSpan = internal.WantSpan{
-		Name:     "http POST unknown",
+		Name:     "http POST lambda.us-west-2.amazonaws.com",
 		ParentID: internal.MatchAnyParent,
 		Attributes: map[string]interface{}{
 			"sampled":       true,
@@ -98,7 +98,7 @@ var (
 		},
 	}
 	externalSpanNoRequestID = internal.WantSpan{
-		Name:     "http POST unknown",
+		Name:     "http POST lambda.us-west-2.amazonaws.com",
 		ParentID: internal.MatchAnyParent,
 		Attributes: map[string]interface{}{
 			"sampled":       true,
--- a/v4/integrations/nrgrpc/nrgrpc_server_test.go
+++ b/v4/integrations/nrgrpc/nrgrpc_server_test.go
@@ -108,7 +108,7 @@ func TestUnaryServerInterceptor(t *testing.T) {
 		},
 		{
 			Name:     "TestApplication/DoUnaryUnary",
-			ParentID: internal.MatchNoParent,
+			ParentID: internal.MatchAnyParent,
 			Attributes: map[string]interface{}{
 				"category": "generic",
 				"parentId": internal.MatchAnything,
@@ -278,7 +278,7 @@ func TestUnaryStreamServerInterceptor(t *testing.T) {
 		},
 		{
 			Name:     "TestApplication/DoUnaryStream",
-			ParentID: internal.MatchNoParent,
+			ParentID: internal.MatchAnyParent,
 			Attributes: map[string]interface{}{
 				"category": "generic",
 				"parentId": internal.MatchAnything,
@@ -377,7 +377,7 @@ func TestStreamUnaryServerInterceptor(t *testing.T) {
 		},
 		{
 			Name:     "TestApplication/DoStreamUnary",
-			ParentID: internal.MatchNoParent,
+			ParentID: internal.MatchAnyParent,
 			Attributes: map[string]interface{}{
 				"category": "generic",
 				"parentId": internal.MatchAnything,
@@ -489,7 +489,7 @@ func TestStreamStreamServerInterceptor(t *testing.T) {
 		},
 		{
 			Name:     "TestApplication/DoStreamStream",
-			ParentID: internal.MatchNoParent,
+			ParentID: internal.MatchAnyParent,
 			Attributes: map[string]interface{}{
 				"category": "generic",
 				"parentId": internal.MatchAnything,
--- a/v4/integrations/nrmicro/nrmicro_test.go
+++ b/v4/integrations/nrmicro/nrmicro_test.go
@@ -600,7 +600,7 @@ func TestServerWrapperWithApp(t *testing.T) {
 		},
 		{
 			Name:     "TestHandler.Method",
-			ParentID: internal.MatchNoParent,
+			ParentID: internal.MatchAnyParent,
 			Attributes: map[string]interface{}{
 				"category":                      "generic",
 				"transaction.name":              "WebTransaction/Go/TestHandler.Method",
--- a/v4/newrelic/segments.go
+++ b/v4/newrelic/segments.go
@@ -259,9 +259,10 @@ func (s *ExternalSegment) url() (*url.URL, error) {
 
 func (s *ExternalSegment) host() string {
 	host := s.Host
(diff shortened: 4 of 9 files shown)


v4/integrations/nrawssdk-v1/nrawssdk_test.go
```diff
@@ -98,7 +98,7 @@ var (
 		},
 	}
 	externalSpanNoRequestID = internal.WantSpan{
-		Name:     "http POST unknown",
+		Name:     "http POST lambda.us-west-2.amazonaws.com",
```
purple4reina, line 101: #### Bug 1

This bug was uncovered by the `External` metrics. In those we were specifically looking for this host. That made me realize that we _should_ be collecting the host, but were not. This was because the host was set in the `Request.URL` which we were not checking as part of creating the span name.

v4/newrelic/segments.go
```diff
@@ -243,13 +243,25 @@ func (s *ExternalSegment) library() string {
...
 func (s *ExternalSegment) host() string {
 	host := s.Host
-	if host == "" && s.URL != "" {
-		url, err := url.Parse(s.URL)
-		if err == nil {
-			host = url.Host
-		}
+	url, err := s.url()
+	if err == nil && host == "" && url.Host != "" {
+		host = url.Host
```
purple4reina, line 264: #### Bug 1

The host should be looked for in the `s.Host`, `s.URL`, _and_ the `s.Request` and `s.Response`. We were not yet checking for it on the request/response.

v4/integrations/nrgrpc/nrgrpc_server_test.go
```diff
@@ -108,7 +108,7 @@ func TestUnaryServerInterceptor(t *testing.T) {
 		},
 		{
 			Name:     "TestApplication/DoUnaryUnary",
-			ParentID: internal.MatchNoParent,
+			ParentID: internal.MatchAnyParent,
```
purple4reina, line 111: #### Bug 2

When a distributed tracing payload is successfully received when calling `txn.AcceptDistributedTraceHeaders`, we created a `TransportDuration` metric. Therefore, the presence of this metric would indicate that _every_ span in the transaction should have a parent.

This lead me to uncover that we should be calling `AcceptDistributedTraceHeaders` in `SetWebRequest`. We haven't technically implemented that method, but I did add the call to Accept in it in order to improve our testing.

v4/newrelic/transaction.go
```diff
@@ -116,7 +116,9 @@ func (txn *Transaction) SetWebRequestHTTP(r *http.Request) {}
 // these fields are set.  If headers are present, the agent will look for
 // distributed tracing headers using Transaction.AcceptDistributedTraceHeaders.
 // Use Transaction.SetWebRequestHTTP if you have a *http.Request.
-func (txn *Transaction) SetWebRequest(r WebRequest) {}
+func (txn *Transaction) SetWebRequest(r WebRequest) {
+	txn.AcceptDistributedTraceHeaders(TransportHTTP, r.Header)
+}
```
purple4reina, line 121: #### Bug 2

We should be calling Accept when the transaction is set to a web request.

v4/newrelic/segments.go
```diff
@@ -301,6 +313,9 @@ func (s *MessageProducerSegment) name() string {
 	if s.DestinationTemporary {
 		dest = "(temporary)"
 	}
+	if dest == "" {
+		dest = "Unknown"
+	}
```
purple4reina, line 318: #### Bug 3

Based on a `MessageBroker` metric, I realized that we should be setting the destination to `Unknown` when it is unknown.

v4/newrelic/segments_test.go
```diff
@@ -510,6 +511,33 @@ func TestExternalSegmentNaming(t *testing.T) {
...
+				Response: &http.Response{
+					Request: &http.Request{
+						URL: &url.URL{
+							Host: "responseHost",
+						},
+					},
+				},
+			},
+			name: "http GET responseHost",
+		},
```
purple4reina, line 540: Tests for **Bug 1**

v4/newrelic/segments.go
```diff
@@ -243,11 +243,24 @@ func (s *ExternalSegment) library() string {
...
+	return nil, nil
+}
+
 func (s *ExternalSegment) host() string {
 	host := s.Host
-	if host == "" && s.URL != "" {
-		url, err := url.Parse(s.URL)
-		if err == nil {
+	if host == "" {
+		if url, _ := s.url(); url != nil {
```
purple4reina, line 263: #### Bug 1
The host should be looked for in the s.Host, s.URL, and the s.Request and s.Response. We were not yet checking for it on the request/response.

v4/newrelic/segments_test.go
```diff
@@ -618,6 +646,12 @@ func TestMessageProducerSegmentNaming(t *testing.T) {
 			},
 			name: "(temporary) send",
 		},
+		{
+			seg: &MessageProducerSegment{
+				DestinationName: "",
+			},
+			name: "Unknown send",
+		},
```
purple4reina, line 654: Tests for **Bug 2**

v4/integrations/nrmicro/nrmicro_test.go
```diff
@@ -600,7 +600,7 @@ func TestServerWrapperWithApp(t *testing.T) {
 		},
 		{
 			Name:     "TestHandler.Method",
-			ParentID: internal.MatchNoParent,
+			ParentID: internal.MatchAnyParent,
```
purple4reina, line 603: #### Bug 2

v4/newrelic/transaction_test.go
```diff
@@ -257,3 +257,56 @@ func TestInsertDistributedTraceHeadersB3(t *testing.T) {
...
+
+	if txnTraceID != remoteTraceID {
+		t.Errorf("txn does not have remote trace id: txnTraceID=%s, remoteTraceID=%s",
+			txnTraceID, remoteTraceID)
+	}
+	if txnParentID != remoteSpanID {
+		t.Errorf("txn is not a child of remote segment: txnParentID=%s, remoteSpanID=%s",
+			txnParentID, remoteSpanID)
+	}
+}
```
purple4reina, line 312: Tests for **Bug 2**

v4/newrelic/transaction.go
```diff
@@ -109,14 +110,39 @@ func (txn *Transaction) AddAttribute(key string, value interface{}) {}
...
 // these fields are set.  If headers are present, the agent will look for
 // distributed tracing headers using Transaction.AcceptDistributedTraceHeaders.
 // Use Transaction.SetWebRequestHTTP if you have a *http.Request.
-func (txn *Transaction) SetWebRequest(r WebRequest) {}
+func (txn *Transaction) SetWebRequest(r WebRequest) {
+	txn.AcceptDistributedTraceHeaders(r.Transport, r.Header)
+}
+
+func transport(r *http.Request) TransportType {
+	if strings.HasPrefix(r.Proto, "HTTP") {
```
pyohannes, line 138: I thought `Proto` can anyway only hold either `HTTP/1.1` or `HTTP/2`. But it doesn't harm to be cautious.

purple4reina: Yeah, I'm completely unsure of this. I just copied this method from the `v3` agent, to keep things consistent when people upgrade. Plus, at this point, we don't even use the `Transport` argument sent to `AcceptDistributedTraceHeaders`. I debated leaving this out entirely for that reason, but decided to keep it around to make things easier on our future selves.

PR #205, opened 2020-08-10
Support custom attributes on segments and transactions.

purple4reina:
Implements all `*.AddAttribute` APIs and tests to make sure those APIs work as expected.

What is not included here is testing for attributes in integration packages. Because I wished to use the `app.ExpectSpanEvents` method for testing, I had to find a way to state specifically when I wanted attributes to be tested and when not. This is why I added the `SkipAttrsTest` field to `WantSpan`. Once all transaction and segment level attributes (ie we've completed work to add "intrinsic" and "agent" attributes) we should remove this field.

### DIFF
--- a/v4/internal/integrationsupport/integrationsupport_test.go
+++ b/v4/internal/integrationsupport/integrationsupport_test.go
@@ -36,8 +36,9 @@ func TestSuccess(t *testing.T) {
 
 	app.ExpectSpanEvents(t, []internal.WantSpan{
 		{
-			Name:     "mySegment",
-			ParentID: internal.MatchAnyParent,
+			Name:          "mySegment",
+			ParentID:      internal.MatchAnyParent,
+			SkipAttrsTest: true,
 			Attributes: map[string]interface{}{
 				"parentId":                         internal.MatchAnything,
 				"category":                         "generic",
@@ -45,8 +46,9 @@ func TestSuccess(t *testing.T) {
 			},
 		},
 		{
-			Name:     "hello",
-			ParentID: internal.MatchNoParent,
+			Name:          "hello",
+			ParentID:      internal.MatchNoParent,
+			SkipAttrsTest: true,
 			Attributes: map[string]interface{}{
 				"transaction.name": "OtherTransaction/Go/hello",
 				"category":         "generic",
--- a/v4/newrelic/segments.go
+++ b/v4/newrelic/segments.go
@@ -170,7 +170,15 @@ func (s *span) isEnded() bool {
 //
 // The key must contain fewer than than 255 bytes.  The value must be a
 // number, string, or boolean.
-func (s *Segment) AddAttribute(key string, val interface{}) {}
+func (s *Segment) AddAttribute(key string, val interface{}) {
+	if s == nil {
+		return
+	}
+	if s.StartTime.span == nil {
+		return
+	}
+	s.StartTime.Span.SetAttribute(key, val)
+}
 
 // End finishes the segment.
 func (s *Segment) End() {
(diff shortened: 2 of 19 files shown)


v4/newrelic/transaction.go
```diff
@@ -49,8 +49,8 @@ func (txn *Transaction) End() {
 	if txn.isEnded() {
 		return
 	}
-	txn.rootSpan.end()
 	txn.thread.Lock()
+	txn.rootSpan.end()
```
purple4reina, line 53: Now I'm second guessing this change. Is it really necessary to end the span under lock?

v4/internal/integrationsupport/integrationsupport_test.go
```diff
@@ -36,17 +36,19 @@ func TestSuccess(t *testing.T) {
...
 				"category":                         "generic",
 				newrelic.SpanAttributeAWSOperation: "operation",
 			},
 		},
 		{
-			Name:     "hello",
-			ParentID: internal.MatchNoParent,
+			Name:          "hello",
+			ParentID:      internal.MatchNoParent,
+			SkipAttrsTest: true,
```
newrelic-eheinlein, line 51: It looks like we're setting `SkipAttrsTest` to `true` a lot. Would it make sense to change it to `IncludeAttrsTest` and have it default to false? Would that be less verbose?

newrelic-eheinlein: Wait, never mind, I didn't read this part:
> The field should be removed once all segment and transaction level attributes have been added.

v4/newrelic/segments.go
```diff
@@ -170,7 +170,15 @@ func (s *span) isEnded() bool {
...
 // number, string, or boolean.
-func (s *Segment) AddAttribute(key string, val interface{}) {}
+func (s *Segment) AddAttribute(key string, val interface{}) {
+	if s == nil {
+		return
+	}
+	if s.StartTime.span == nil {
+		return
+	}
+	s.StartTime.Span.SetAttribute(key, val)
```
newrelic-eheinlein, line 180: This doesn't actually have to do with this PR, it's just something that's been floating around in my head: I don't like the name `StartTime`.  It seems counter intuitive to what it actually represents. I'm guessing we need to keep it that way because it is public and we don't want to change the API, but if there is any way as we work on this shim work that we can find to change it to something a little more intuitive, I would like that.

v4/newrelic/segments.go
```diff
@@ -295,7 +319,15 @@ func (s *ExternalSegment) method() string {
 //
 // The key must contain fewer than than 255 bytes.  The value must be a
 // number, string, or boolean.
-func (s *MessageProducerSegment) AddAttribute(key string, val interface{}) {}
+func (s *MessageProducerSegment) AddAttribute(key string, val interface{}) {
```
newrelic-eheinlein, line 322: This does feel like a lot of repeated code.  Is there any way we can make the different segment types inherit from a single `Segement` type, maybe via embedding it, or something? I don't know if that would break anything, but it might allow us to write this `AddAttribute` code only once.

v4/newrelic/transaction.go
newrelic-eheinlein: After an initial look, it seems to me that the locking is important when setting the root span in `setCurrentSpan`, but it does the locking there:

```go
func (thd *thread) setCurrentSpan(s *span) {
	thd.Lock()
	thd.currentSpan = s
	thd.isMultiSpan = true
	thd.Unlock()
}
```
So my first thought is that it shouldn't be necessary, or in fact could maybe be problematic? But I'll think about it some more before I am sure of that. Might be good to write some tests around this if possible.

newrelic-eheinlein:
I posted a couple of thoughts, but they are more broad and might not be good for this specific PR.  Maybe we can chat about them in person (well, as in person as things get these days).

v4/newrelic/transaction.go
purple4reina: Yeah, I think I'll change this back. Thanks for the sleuthing!

v4/newrelic/segments.go
purple4reina: Agreed, I've been feeling the same thing. But yeah, the problem is that it's public so we can't change the name.

v4/newrelic/segments.go
purple4reina: The imbedded type would need to have access to the `StartTime`, which means that field I think would need to be on the imbedded type, and that's not possible because its changing the API. instead, we could create a single method that all of these `AddAttribute` methods call into, but even that seemed heavy handed to me. I decided to go with this because it was very explicit. If you wanted to know how `AddAttribute` works, you don't need to jump around to a million places just to figure it out. It ends up being a line or two more code this way in total, but ultimately, I think it's the better option for readability.

PR #206, opened 2020-08-11
Add attributes to spans created from DatastoreSegments.

purple4reina:
I've left comments throughout the code to aid in your review. Please take a good look at the specification for open telemetry datastore spans [here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md).


v4/integrations/nrawssdk-v1/nrawssdk_test.go
```diff
@@ -158,14 +158,14 @@ var (
...
 		{Name: "Datastore/all", Scope: "", Forced: true, Data: nil},
 		{Name: "Datastore/allOther", Scope: "", Forced: true, Data: nil},
-		{Name: "Datastore/instance/DynamoDB/dynamodb.us-west-2.amazonaws.com/unknown", Scope: "", Forced: false, Data: nil},
-		{Name: "Datastore/operation/DynamoDB/DescribeTable", Scope: "", Forced: false, Data: nil},
-		{Name: "Datastore/statement/DynamoDB/thebesttable/DescribeTable", Scope: "", Forced: false, Data: nil},
-		{Name: "Datastore/statement/DynamoDB/thebesttable/DescribeTable", Scope: "OtherTransaction/Go/" + txnName, Forced: false, Data: nil},
+		{Name: "Datastore/instance/dynamodb/dynamodb.us-west-2.amazonaws.com/unknown", Scope: "", Forced: false, Data: nil},
+		{Name: "Datastore/operation/dynamodb/DescribeTable", Scope: "", Forced: false, Data: nil},
+		{Name: "Datastore/statement/dynamodb/thebesttable/DescribeTable", Scope: "", Forced: false, Data: nil},
+		{Name: "Datastore/statement/dynamodb/thebesttable/DescribeTable", Scope: "OtherTransaction/Go/" + txnName, Forced: false, Data: nil},
```
purple4reina, line 168: `dynamodb` is now lower cased in all the tests because the otel spec wants it that way.

v4/integrations/nrelasticsearch-v7/nrelastic_test.go
```diff
@@ -495,10 +495,10 @@ func TestInfoRequest(t *testing.T) {
...
 		{Name: "Datastore/all", Scope: "", Forced: nil, Data: nil},
 		{Name: "Datastore/allOther", Scope: "", Forced: nil, Data: nil},
-		{Name: "Datastore/Elasticsearch/all", Scope: "", Forced: nil, Data: nil},
-		{Name: "Datastore/Elasticsearch/allOther", Scope: "", Forced: nil, Data: nil},
-		{Name: "Datastore/operation/Elasticsearch/info", Scope: "", Forced: nil, Data: nil},
-		{Name: "Datastore/operation/Elasticsearch/info", Scope: "OtherTransaction/Go/txnName", Forced: nil, Data: nil},
+		{Name: "Datastore/elasticsearch/all", Scope: "", Forced: nil, Data: nil},
+		{Name: "Datastore/elasticsearch/allOther", Scope: "", Forced: nil, Data: nil},
+		{Name: "Datastore/operation/elasticsearch/info", Scope: "", Forced: nil, Data: nil},
+		{Name: "Datastore/operation/elasticsearch/info", Scope: "OtherTransaction/Go/txnName", Forced: nil, Data: nil},
```
purple4reina, line 501: Otel wants the `e` in Elasticsearch to be lower case.

v4/integrations/nrmongo/nrmongo_test.go
```diff
@@ -196,19 +196,19 @@ func TestMonitor(t *testing.T) {
...
+		{Name: "Datastore/mongodb/all", Scope: "", Forced: true, Data: []float64{2.0}},
+		{Name: "Datastore/mongodb/allOther", Scope: "", Forced: true, Data: []float64{2.0}},
 		{Name: "OtherTransaction/all", Scope: "", Forced: true, Data: nil},
 		{Name: "OtherTransactionTotalTime", Scope: "", Forced: true, Data: nil},
 		{Name: "DurationByCaller/Unknown/Unknown/Unknown/Unknown/all", Scope: "", Forced: false, Data: nil},
 		{Name: "DurationByCaller/Unknown/Unknown/Unknown/Unknown/allOther", Scope: "", Forced: false, Data: nil},
-		{Name: "Datastore/statement/MongoDB/collName/commName", Scope: "", Forced: false, Data: []float64{2.0}},
-		{Name: "Datastore/statement/MongoDB/collName/commName", Scope: "OtherTransaction/Go/txnName", Forced: false, Data: []float64{2.0}},
+		{Name: "Datastore/statement/mongodb/collName/commName", Scope: "", Forced: false, Data: []float64{2.0}},
+		{Name: "Datastore/statement/mongodb/collName/commName", Scope: "OtherTransaction/Go/txnName", Forced: false, Data: []float64{2.0}},
```
purple4reina, line 211: Similar here, mongodb all lower case.

v4/integrations/nrnats/test/nrnats_test.go
```diff
@@ -186,7 +186,7 @@ func TestStartPublishSegmentNaming(t *testing.T) {
 		subject string
 		metric  string
 	}{
-		{subject: "", metric: "MessageBroker/NATS/Topic/Produce/Named/Unknown"},
+		{subject: "", metric: "MessageBroker/NATS/Topic/Produce/Named/unknown"},
```
purple4reina, line 189: This needed to be updated because I made all the `"unknown"`s consistent to use the same method. This means they all need to be lowercased.

v4/newrelic/datastore.go
```diff
@@ -9,26 +9,26 @@ type DatastoreProduct string
...
+	DatastoreNeptune       DatastoreProduct = "neptune"
+	DatastoreOracle        DatastoreProduct = "oracle"
+	DatastorePostgres      DatastoreProduct = "postgresql"
+	DatastoreRedis         DatastoreProduct = "redis"
+	DatastoreRiak          DatastoreProduct = "riak"
+	DatastoreSQLite        DatastoreProduct = "sqlite"
+	DatastoreSnowflake     DatastoreProduct = "snowflake"
+	DatastoreSolr          DatastoreProduct = "solr"
+	DatastoreTarantool     DatastoreProduct = "tarantool"
+	DatastoreVoltDB        DatastoreProduct = "voltDB"
```
purple4reina, line 33: The otel spec is very clear about the names for each of these: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#notes-and-well-known-identifiers-for-dbsystem

v4/newrelic/segments.go
```diff
@@ -214,18 +216,58 @@ func (s *DatastoreSegment) End() {
 	if s.StartTime.isEnded() {
 		return
 	}
+
+	s.addAttributes(s.StartTime.Span.SetAttribute)
 	s.StartTime.Span.SetName(s.name())
 	s.StartTime.end()
 }
 
+func (s *DatastoreSegment) addAttributes(set func(string, interface{})) {
```
purple4reina, line 225: Passing a `set` func here as an arg made testing easier. It also made the code easier to read.

v4/newrelic/segments.go
```diff
@@ -214,18 +216,58 @@ func (s *DatastoreSegment) End() {
...
+	case DatastoreCassandra:
+		set("db.cassandra.keyspace", valOrUnknown(s.DatabaseName))
+	case DatastoreRedis:
+		set("db.redis.database_index", valOrUnknown(s.DatabaseName))
+	case DatastoreMongoDB:
+		set("db.mongodb.collection", valOrUnknown(s.DatabaseName))
+	default:
+		set("db.name", valOrUnknown(s.DatabaseName))
+	}
+}
```
purple4reina, line 251: Please check the otel spec to make sure I got all my logic correct. https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md

v4/newrelic/segments.go
```diff
@@ -214,18 +216,58 @@ func (s *DatastoreSegment) End() {
...
+	set("db.statement", valOrUnknown(s.statement()))
+	set("db.operation", valOrUnknown(s.Operation))
+
+	if net.ParseIP(s.Host) != nil {
+		set("net.peer.ip", s.Host)
+	} else {
+		set("net.peer.name", valOrUnknown(s.Host))
+	}
+	if s.PortPathOrID != "" {
+		if port, err := strconv.Atoi(s.PortPathOrID); err == nil {
```
purple4reina, line 236: They specifically want the port attribute to be an int.

v4/newrelic/segments.go
```diff
@@ -214,18 +216,58 @@ func (s *DatastoreSegment) End() {
...
+	default:
+		set("db.name", valOrUnknown(s.DatabaseName))
+	}
+}
+
 func (s *DatastoreSegment) name() string {
+	return s.statement()
+}
+
+func valOrUnknown(v string) string {
```
purple4reina, line 257: I made this func so we could use it throughout this file. I made several replacements using it.

v4/newrelic/segments.go
```diff
@@ -214,18 +216,58 @@ func (s *DatastoreSegment) End() {
...
 	if pq == "" {
-		coll := s.Collection
-		if "" == coll {
-			coll = "unknown"
-		}
-		pq = "'" + s.Operation + "' on '" + coll + "' using '" + string(s.Product) + "'"
+		op := valOrUnknown(s.Operation)
+		coll := valOrUnknown(s.Collection)
+		prod := valOrUnknown(string(s.Product))
+		pq = "'" + op + "' on '" + coll + "' using '" + prod + "'"
```
purple4reina, line 270: This change updated the names of several spans in our tests. I feel like using `'unknown'` instead of `''` in places is easier to read and understand.

v4/newrelic/segments_test.go
```diff
@@ -410,7 +410,7 @@ func TestDatastoreSegmentNaming(t *testing.T) {
 				PortPathOrID: "port",
 				DatabaseName: "dbname",
 			},
-			name: "'operation' on 'collection' using 'Postgres'",
+			name: "'operation' on 'collection' using 'postgresql'",
```
purple4reina, line 413: The otel spec is very clear about how this should look.

v4/newrelic/segments_test.go
```diff
@@ -425,7 +425,11 @@ func TestDatastoreSegmentNaming(t *testing.T) {
 				PortPathOrID: "port",
 				DatabaseName: "dbname",
 			},
-			name: "'operation' on 'unknown' using 'Postgres'",
+			name: "'operation' on 'unknown' using 'postgresql'",
+		},
+		{
+			seg:  &DatastoreSegment{},
+			name: "'unknown' on 'unknown' using 'unknown'",
```
purple4reina, line 432: I added this test to cover all the unknowns.

v4/newrelic/segments_test.go
```diff
@@ -622,7 +626,7 @@ func TestMessageProducerSegmentNaming(t *testing.T) {
 			seg: &MessageProducerSegment{
 				DestinationName: "",
 			},
-			name: "Unknown send",
+			name: "unknown send",
```
purple4reina, line 629: This change came with the update to using the `valOrUnknown` method.

v4/newrelic/segments_test.go
```diff
@@ -740,31 +744,49 @@ func TestSegmentsAddAttribute(t *testing.T) {
...
+					StartTime: txn.StartSegmentNow(),
+				}
+			},
+			extraAttrs: map[string]interface{}{
+				"db.system":     "unknown",
+				"db.statement":  "'unknown' on 'unknown' using 'unknown'",
+				"db.operation":  "unknown",
+				"net.peer.name": "unknown",
+				"db.name":       "unknown",
+			},
```
purple4reina, line 768: This block of changes needed to happen because there's now more attributes on some spans that we need to account for. This test is looking for custom attributes on spans. But since the custom attrs and the attrs that our shim makes are all in the same bucket, we must also look for these attrs made by the shim as well.

v4/newrelic/sql_driver.go
```diff
@@ -1,13 +1,12 @@
 // Copyright 2020 New Relic Corporation. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-// +build go1.10
-
```
purple4reina, line 5: v4 only supports v1.13+

v4/newrelic/sql_driver.go
```diff
@@ -187,10 +191,11 @@ func (w *wrapConn) Query(query string, args []driver.Value) (driver.Rows, error)
 
 // QueryContext implements QueryerContext.
 func (w *wrapConn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
-	segment := w.bld.useQuery(query).startSegment(ctx)
+	startTime := time.Now()
 	rows, err := w.original.(driver.QueryerContext).QueryContext(ctx, query, args)
 	if err != driver.ErrSkip {
-		segment.End()
+		seg := w.bld.useQuery(query).startSegmentAt(ctx, startTime)
+		seg.End()
```
purple4reina, line 198: For some reason, this update to the sql driver file wasn't included! I caught this through running some of the tests! In order to add this back in, I also needed to include a `txn.startSegmentAt` method.

v4/newrelic/sql_driver_test.go
```diff
@@ -0,0 +1,295 @@
...
+		{Name: "Datastore/all", Scope: "", Forced: true, Data: nil},
+		{Name: "Datastore/allOther", Scope: "", Forced: true, Data: nil},
+		{Name: "Datastore/mysql/all", Scope: "", Forced: true, Data: nil},
+		{Name: "Datastore/mysql/allOther", Scope: "", Forced: true, Data: nil},
+		{Name: "Datastore/operation/mysql/myoperation", Scope: "", Forced: false, Data: nil},
+		{Name: "Datastore/statement/mysql/mycollection/myoperation", Scope: "", Forced: false, Data: nil},
+		{Name: "Datastore/statement/mysql/mycollection/myoperation", Scope: "OtherTransaction/Go/hello", Forced: false, Data: nil},
+		{Name: "Datastore/instance/mysql/myhost/myport", Scope: "", Forced: false, Data: nil},
+	}
+	driverTestSpans = []internal.WantSpan{
```
purple4reina, line 30: This file was just copied from the v3 directory. In addition, I added testing for spans as well, since this file in the v3 directory only tested metrics.

v4/newrelic/sql_driver_test.go
```diff
@@ -0,0 +1,295 @@
...
+	ctx := NewContext(context.Background(), txn)
+	conn.(driver.QueryerContext).QueryContext(ctx, "myoperation,mycollection", nil)
+	txn.StartSegment("second").End()
+	txn.End()
+	app.ExpectSpanEvents(t, []internal.WantSpan{
+		{
+			Name:       "second",
+			SpanID:     "0000000000000003",
+			TraceID:    "00000000000000020000000000000000",
+			ParentID:   "0000000000000002",
```
purple4reina, line 277: It was through updating this test that I discovered we'd forgotten the update to the `sql_driver.go` file! Yay for tests!

v4/newrelic/transaction.go
```diff
@@ -201,7 +206,9 @@ func (txn *Transaction) StartSegmentNow() SegmentStartTime {
 	}
 	parent := txn.thread.getCurrentSpan()
 	ctx, sp := txn.app.tracer.Start(parent.ctx, "",
-		trace.WithSpanKind(trace.SpanKindInternal))
+		trace.WithSpanKind(trace.SpanKindInternal),
+		trace.WithStartTime(at),
+	)
```
purple4reina, line 211: These changes add support for `txn.startSegmentAt`.

v4/integrations/nrawssdk-v1/nrawssdk_test.go
newrelic-eheinlein: For these type of name changes, if they affect anything like alerts or the UI, we should make sure to include that high up in the changelog for this shim.

purple4reina: Since we're not including any metrics in this round of the shim, these changes won't have any specific effect on Alerts. Attributes on spans will definitely be changing, so that's def something we should call out.

v4/newrelic/segments.go
```diff
@@ -198,18 +216,59 @@ func (s *DatastoreSegment) End() {
...
+	s.addAttributes(s.StartTime.Span.SetAttribute)
 	s.StartTime.Span.SetName(s.name())
 	s.StartTime.end()
 }
 
+func (s *DatastoreSegment) addAttributes(set func(string, interface{})) {
+	set("db.system", valOrUnknown(string(s.Product)))
+	set("db.statement", valOrUnknown(s.statement()))
+	set("db.operation", valOrUnknown(s.Operation))
+	set("db.collection", valOrUnknown(s.Collection))
```
purple4reina, line 229: This one attribute is not defined in the opentelemetry specs. I decided to include it because we had this info on our datastore spans in v3 and I didn't want this information to be lost to our customers.

v4/newrelic/segments.go
newrelic-eheinlein: I like this, but I think the name could change.  I see "set" and my first thought was of the data collection type of set, rather than a setter.  Would `setter` be an acceptable name for this parameter to you? 

Then again, I can also accept that I might be abnormal that I thought of a data structure rather than a setter, so if you think that name is better I'm okay with it.

newrelic-eheinlein:
I think I just have one suggestion here, which is the change from `set` to `setter`, and it's just a though, not something I feel strongly about.  Other than that, LGTM!

v4/newrelic/segments.go
purple4reina: Works for me!

PR #207, opened 2020-08-12
Add attributes to spans created from ExternalSegments.

purple4reina:
This pull request adds attributes for external segments. There are only 4 attributes added:

* `http.url`: this is the url for the request
* `http.method`: this is the http method
* `http.status_code`: this is the status code for the response
* `http.component`: this is the value of `ExternalSegment.Library`, OpenTelemetry doesn't call for this attribute specifically but we included this value before so I want to make sure that customers still have access to it

Be sure to read the spec for [http spans](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md) and make sure my logic is correct.

### DIFF
--- a/v4/newrelic/segments.go
+++ b/v4/newrelic/segments.go
@@ -11,7 +11,9 @@ import (
 	"strconv"
 	"sync"
 
+	"go.opentelemetry.io/otel/api/standard"
 	"go.opentelemetry.io/otel/api/trace"
+	"google.golang.org/grpc/codes"
 )
 
 type span struct {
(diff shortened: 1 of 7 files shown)


v4/newrelic/segments.go
```diff
@@ -112,6 +115,10 @@ type ExternalSegment struct {
 	// external metrics and the "component" span attribute.  It should be
 	// the framework making the external call.
 	Library string
+
+	// statusCode is the status code for the response.  This value takes
+	// precedence over the status code set on the Response.
```
newrelic-eheinlein, line 120: This comment is fairly confusing on first read-through, although after I slow down and re-read it a couple of times I understand.  Maybe something like this would be a little more readable?

> statusCode represents the status code that will be reported as the response to this external call.  This value takes precedence over the status code set on the `Response` field.

v4/newrelic/segments.go
```diff
@@ -295,10 +302,74 @@ func (s *ExternalSegment) End() {
...
+	if s.statusCode != nil {
+		code = *s.statusCode
+	} else if s.Response != nil {
+		code = s.Response.StatusCode
+	}
+	if code < 17 {
+		// Assume the code is already a grpc status code
+		c := codes.Code(code)
+		setter(c, c.String())
+		return
```
newrelic-eheinlein, line 322: this is purely stylistic, but I think I would prefer to not `return` here and instead use an `else` block below. I like early `return`s early on in a function when validating preconditions, but this late in the function I find it a little confusing.

purple4reina: No problem! I'll make the change.

PR #210, opened 2020-08-13
Add span attributes for message producer segments.

purple4reina:
Based on the [open telemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md) for messaging. This adds attributes to our `MessageProducerSegment`s.

The specification also says we should be including `net.peer.name`, `net.peer.ip`, and `net.peer.port`. However, our `MessageProducerSegment` does not have fields for these. Adding them would mean adding two new fields onto the segment type plus updating our instrumentation for micro and nats/stan to include these. That is outside the scope of this MMF.

### DIFF
--- a/v4/newrelic/segments.go
+++ b/v4/newrelic/segments.go
@@ -147,9 +147,9 @@ type MessageDestinationType string
 // These message destination type constants are used in for the
 // MessageSegment.DestinationType field.
 const (
-	MessageQueue    MessageDestinationType = "Queue"
-	MessageTopic    MessageDestinationType = "Topic"
-	MessageExchange MessageDestinationType = "Exchange"
+	MessageQueue    MessageDestinationType = "queue"
+	MessageTopic    MessageDestinationType = "topic"
+	MessageExchange MessageDestinationType = "exchange"
 )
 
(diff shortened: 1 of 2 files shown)


v4/newrelic/segments.go
```diff
@@ -378,10 +378,22 @@ func (s *MessageProducerSegment) End() {
 	if s.StartTime.isEnded() {
 		return
 	}
+	s.addAttributes(s.StartTime.Span.SetAttribute)
 	s.StartTime.Span.SetName(s.name())
 	s.StartTime.end()
 }
 
+func (s *MessageProducerSegment) addAttributes(setter func(string, interface{})) {
```
newrelic-eheinlein, line 386: What about the name `addMessageProducerAttrs` or something like that? When I see `addAttributes()` I assume the parameters are going to be the attributes to be added, which isn't the case here. 

(This might affect other segment types too, and I may have just missed it in other PRs if so.)

newrelic-eheinlein:
Just one naming suggestion, plus this has conflicts to be resolved, then it looks good!

PR #211, opened 2020-08-14
Directly add attrs instead of using integrationsupport methods.

purple4reina:
Since there is no longer such things as "agent attributes", these two methods can be removed in favor of the vanilla `AddAttribute` api.

This change will now allow us to begin removing the `SkepAttrsTest` field from our spans tests.


v4/internal/awssupport/awssupport.go
```diff
@@ -10,15 +10,17 @@ import (
...
-	"github.com/newrelic/go-agent/v4/internal/integrationsupport"
 	"github.com/newrelic/go-agent/v4/newrelic"
 )
 
 type contextKeyType struct{}
 
 var segmentContextKey = contextKeyType(struct{}{})
 
-type endable interface{ End() }
+type segmenter interface {
```
newrelic-eheinlein, line 20: 😍

PR #215, opened 2020-08-14
Add logging to the shim

newrelic-eheinlein:
This is currently a WIP. Still to work on ensuring we don't panic when the logger is nil (preferably in a more elegant way than the way I started addressing it)

### DIFF
--- a/v4/go.mod
+++ b/v4/go.mod
@@ -5,4 +5,5 @@ go 1.13
 require (
 	github.com/newrelic/opentelemetry-exporter-go v0.1.0
 	go.opentelemetry.io/otel v0.10.0
+	google.golang.org/grpc v1.30.0
 )
--- a/v4/newrelic/application.go
+++ b/v4/newrelic/application.go
@@ -70,8 +70,8 @@ func (app *Application) RecordCustomEvent(eventType string, params map[string]in
 	if app == nil {
 		return
 	}
-	app.initLogger()
-	logUnimplemented(app.logger, "Application.RecordCustomEvent")
+	app.LogDebug(unimplementedMessage("Application.RecordCustomEvent"), nil)
+
 }
 
 // RecordCustomMetric records a custom metric.  The metric name you
@@ -85,8 +85,7 @@ func (app *Application) RecordCustomMetric(name string, value float64) {
 	if app == nil {
 		return
 	}
-	app.initLogger()
-	logUnimplemented(app.logger, "Application.RecordCustomMetric")
+	app.LogDebug(unimplementedMessage("Application.RecordCustomMetric"), nil)
 }
 
 // WaitForConnection blocks until the application is connected, is
@@ -118,7 +117,7 @@ func (app *Application) Shutdown(timeout time.Duration) {
 	if app == nil {
 		return
 	}
-	logUnimplemented(app.logger, "Application.Shutdown")
+	app.LogDebug(unimplementedMessage("Application.Shutdown"), nil)
 }
 
 // NewApplication creates an Application and spawns goroutines to manage the
@@ -191,9 +190,6 @@ func (app *Application) initLogger() {
 	}
 }
 
-func logUnimplemented(logger Logger, methodName string) {
-	if logger == nil {
-		return
-	}
-	logger.Debug(methodName+" is currently a no-op for this New Relic agent (it may be implemented in a future version)", nil)
+func unimplementedMessage(methodName string) string {
+	return methodName + " is currently a no-op for this New Relic agent (it may be implemented in a future version)"
 }
--- a/v4/newrelic/application.go
+++ b/v4/newrelic/application.go
@@ -102,6 +102,9 @@ func (app *Application) WaitForConnection(timeout time.Duration) error {
 	if app == nil {
 		return nil
 	}
+	if app.logger == nil {
+		app.logger = logger.ShimLogger{}
+	}
 	app.logger.Debug("WaitForConnection is a no-op for this New Relic agent and can be removed", nil)
 	return nil
 }
--- a/v4/newrelic/application.go
+++ b/v4/newrelic/application.go
@@ -11,6 +11,8 @@ import (
 	"go.opentelemetry.io/otel/api/global"
 	"go.opentelemetry.io/otel/api/propagation"
 	"go.opentelemetry.io/otel/api/trace"
+
+	"github.com/newrelic/go-agent/v4/internal/logger"
 )
 
 // Application represents your application.  All methods on Application are nil
@@ -26,13 +28,15 @@ func (app *Application) StartTransaction(name string) *Transaction {
 	if app == nil {
 		return nil
 	}
+	if app.logger == nil {
+		app.logger = logger.ShimLogger{}
+	}
 	if app.tracer == nil {
-		if app.logger != nil {
-			app.logger.Debug(
-				"trying to start a transaction, but the OpenTelemetry.Tracer is not set in the config; aborting",
-				nil,
-			)
-		}
+		app.logger.Debug(
+			"trying to start a transaction, but the OpenTelemetry.Tracer is not set in the config; aborting",
+			nil,
+		)
+
 		return nil
 	}
 	ctx, sp := app.tracer.Start(context.Background(), name,
@@ -98,9 +102,7 @@ func (app *Application) WaitForConnection(timeout time.Duration) error {
 	if app == nil {
 		return nil
 	}
-	if app.logger != nil {
-		app.logger.Debug("WaitForConnection is a no-op for this New Relic agent and can be removed", nil)
-	}
+	app.logger.Debug("WaitForConnection is a no-op for this New Relic agent and can be removed", nil)
 	return nil
 }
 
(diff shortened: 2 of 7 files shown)


v4/newrelic/application.go
```diff
@@ -5,26 +5,36 @@ package newrelic
...
 	propagators propagation.Propagators
+	logger      Logger
 }
 
 // StartTransaction begins a Transaction with the given name.
 func (app *Application) StartTransaction(name string) *Transaction {
 	if app == nil {
 		return nil
 	}
+	app.initLogger()
```
purple4reina, line 31: This should be in `NewApplication`.

v4/newrelic/application.go
```diff
@@ -76,6 +106,11 @@ func (app *Application) RecordCustomMetric(name string, value float64) {}
 // connection to the Trace Observer is made, a fatal error is reached, or the
 // timeout is hit.
 func (app *Application) WaitForConnection(timeout time.Duration) error {
+	if app == nil {
+		return nil
+	}
+	app.initLogger()
```
purple4reina, line 112: We should just put this in `NewApplication` so we don't need to worry about it here.

v4/newrelic/application.go
newrelic-eheinlein: That was where I had it originally, but tests were failing if someone just creates an app manually - this test:

v4/newrelic/application.go
```diff
@@ -55,7 +66,13 @@ func (app *Application) StartTransaction(name string) *Transaction {
 // https://docs.newrelic.com/docs/insights/new-relic-insights/adding-querying-data/inserting-custom-events-new-relic-apm-agents
 //
 // An error is logged if eventType or params is invalid.
-func (app *Application) RecordCustomEvent(eventType string, params map[string]interface{}) {}
+func (app *Application) RecordCustomEvent(eventType string, params map[string]interface{}) {
+	if app == nil {
+		return
+	}
+	app.initLogger()
```
joshuabenuck, line 73: Why is `app.initLogger()` being called here when it isn't being called before `logUnimplemented` in `app.Shutdown`?

v4/newrelic/application.go
```diff
@@ -120,5 +151,49 @@ func NewApplication(opts ...ConfigOption) (*Application, error) {
...
+	app.logger.Debug(msg, context)
+}
+
+func (app *Application) initLogger() {
+	if app.logger == nil {
+		app.logger = logger.ShimLogger{}
+	}
+}
+
+func logUnimplemented(logger Logger, methodName string) {
```
joshuabenuck, line 194: Why is this not a method on `app` like the other logging methods?

v4/go.mod
```diff
@@ -5,4 +5,5 @@ go 1.13
 require (
 	github.com/newrelic/opentelemetry-exporter-go v0.1.0
 	go.opentelemetry.io/otel v0.10.0
+	google.golang.org/grpc v1.30.0
```
joshuabenuck, line 8: What led to this being added? It isn't immediately apparent in the diffs of the other files.

v4/newrelic/application.go
newrelic-eheinlein: The answer to this one is also the answer to the previous question. Long story short, I was trying to repeat code as little as possible, and I used this `logUnimplemented` method in a few different places, including on the `Application`, but also on the `Transaction`. I didn't want to check the `app.logger` more times than I need to. But since you brought up the question I am realizing that this isn't very readable and is probably more prone to bugs. I'll try to improve it.

joshuabenuck: Not sure if this is where you are headed with your refactor... Mentioning it to provide extra input in case it helps. When reading through the logging changes, I am left wondering why the majority of the logging functions are being added to `Application` instead of `Logger`? All the call sites seem to have access to a reference to a `Logger`. Not all of them have access to an `Application`.

newrelic-eheinlein: Originally I added them to the `Logger`, but the issue I ran into was that someone could conceivably manually create an `Application` object (rather than calling `NewApplication`), in which case the `Logger` on the app would be `nil`, which could lead to panics if we don't do a nil check each time. The idea was to add methods on the app so that we could keep the nil checks all in one place. I think I need to make sure to follow a similar pattern with `Transaction` so that it's less confusing, or at least that's what I'm trying first to see how it looks.

v4/go.mod
newrelic-eheinlein: This should be from the upgrade of otel to 0.1.0 I believe.

v4/newrelic/segments.go
```diff
@@ -196,12 +196,21 @@ func (s *Segment) End() {
...
 	if s.StartTime.isEnded() {
+		logSgmtAlreadyEnded(s.StartTime, s.Name)
 		return
 	}
 	s.StartTime.Span.SetName(s.Name)
 	s.StartTime.end()
 }
 
+func logSgmtAlreadyEnded(s SegmentStartTime, name string) {
+	if s.span != nil && s.thread != nil {
```
joshuabenuck, line 207: Do we need to add a `s.thread.logger != nil` clause?

v4/newrelic/segments.go
```diff
@@ -196,6 +201,7 @@ func (s *Segment) End() {
 		return
 	}
 	if s.StartTime.isEnded() {
+		logAlreadyEnded(s.StartTime, s.Name)
```
newrelic-eheinlein, line 204: A note on why I use a function here rather than just adding a logging method on the segment: there are a bunch of different types of segments, and I didn't want to add a method to each one. Also on each of these segment types the logger is pretty nested (`s.StartTime.span.thread.logger`), and most of those things along the path can be `nil` but I didn't want to check that whole path explicitly for every segment type.

Each segment type has a `SegmentStartTime` field which can't be `nil` because it's a struct, so that's why I chose `s.StartTime` to pass in to the method.

v4/newrelic/segments.go
```diff
@@ -448,13 +456,22 @@ func (s *MessageProducerSegment) End() {
...
+		logAlreadyEnded(s.StartTime, s.name())
 		return
 	}
 	s.addRequiredAttributes(s.StartTime.Span.SetAttribute)
 	s.StartTime.Span.SetName(s.name())
 	s.StartTime.end()
 }
 
+func logAlreadyEnded(st SegmentStartTime, name string) {
+	if st.span != nil {
```
newrelic-eheinlein, line 468: the path to the logger from here is `st.span.thread.logger`, but calling `st.thread.logDebug` handles the check for the nil receiver (where the receiver is the thread) so therefore here I am only checking the `st.span` for nil.

v4/newrelic/transaction.go
```diff
@@ -35,6 +35,23 @@ type thread struct {
 	// isMultiSpan is set to true of the associated transaction has more
 	// than one span or crosses multiple goroutines.
 	isMultiSpan bool
+	logger      Logger
+}
+
+func (thd *thread) logDebug(msg string, context map[string]interface{}) {
```
newrelic-eheinlein, line 41: I chose to add these `logDebug` and `logError` methods to the thread, rather than the transaction, because the thread is the common structure shared between `Transaction` and each of the `Segment` types and I wanted to be able to reuse the methods. I feel like it would be cleaner to read to add these methods directly to `Transaction` and each of the segment types, but I didn't want to copy the code so many times.  So to me this was a trade-off between readability and code reuse, and I went for code reuse. Happy to debate this choice, though.

v4/newrelic/log.go
```diff
@@ -26,18 +25,12 @@ type Logger interface {
 	DebugEnabled() bool
 }
 
-// NewLogger creates a basic Logger at info level.
-//
-// Deprecated: NewLogger is deprecated and will be removed in a future release.
-// Use the ConfigInfoLogger ConfigOption instead.
-func NewLogger(w io.Writer) Logger {
+// newLogger creates a basic Logger at info level.
+func newLogger(w io.Writer) Logger {
```
newrelic-eheinlein, line 29: I changed this to non-exported because it was already deprecated in the v3 agent.

v4/newrelic/transaction.go
```diff
@@ -63,11 +80,19 @@ func (txn *Transaction) isEnded() bool {
 }
 
 // Ignore prevents this transaction's data from being recorded.
-func (txn *Transaction) Ignore() {}
+func (txn *Transaction) Ignore() {
+	if txn != nil && txn.thread != nil {
```
newrelic-eheinlein, line 84: I went back and forth between having these two checks in a single method (`txn.canLog()`) and I can go either way on it. We check it enough that it makes sense to have it in its own method to me, but at some point I felt differently and I haven't changed it back. 🤷‍♀️

joshuabenuck: You have a `thd.canLog()`, but not a `txn.canLog()`. I think I'm slightly inclined to have `canLog()`'s as they are DRY'er. I can go either way, but I do think we should be consistent.

v4/newrelic/application.go
```diff
@@ -120,5 +145,42 @@ func NewApplication(opts ...ConfigOption) (*Application, error) {
...
+}
+
+func (app *Application) logInfo(msg string, context map[string]interface{}) {
+	if app != nil {
+		app.initLogger()
+		app.logger.Info(msg, context)
+	}
+}
+
+func (app *Application) logDebug(msg string, context map[string]interface{}) {
```
joshuabenuck, line 171: This is now not exported. Did you decide against having these logging methods exposed to users of the API? I'd be interested in what you found that led to the reversal. (There's a definite advantage to not exporting them assuming we are confident that their absence will not be blocking to any users of the agent).

v4/newrelic/segments.go
joshuabenuck: Makes sense. If these were exported fields it would matter a whole lot more about where store the logger reference (and whether we use a function here vs add methods to the segment). Thanks for checking on whether StartTime can be `nil`!

v4/newrelic/segments.go
joshuabenuck: If the path to the logger is `st.span.thread.logger`, why isn't the call `st.span.thread.logDebug`?

newrelic-eheinlein: In this case the `span` is embedded in the `SegmentStartTime` so you can call `st.thread` directly, however if the `span` is nil it will still panic (even though it's not explicitly in the call).

v4/newrelic/application.go
newrelic-eheinlein: Yep, I took a look at the v3 agent and realized that we don't expose any logging to the customer. If they want to log something, they can do it in their own way; no need to expose a way for them to log specifically via the agent. I don't remember why I thought it should be exported to start with, but it does not appear to be necessary.

v4/newrelic/transaction.go
newrelic-eheinlein: Cool, I agree - I'll change it back to better reuse the code.

PR #216, opened 2020-08-14
Enable testing for all spans created from segments.

purple4reina:
NOTE: This pull request relies #214 and the update of localhost attributes on datastore spans.

This pull request removes the `SkipAttrsTest` field from most spans testing. All spans that were created from segments (ie all those except the ones created from `app.StartTransaction`) now have the `SkipAttrsTest` field removed.

I also went ahead and lowercased `grpc` in our integration package. It seems that OpenTelemetry prefers all lowercase names.


v4/newrelic/segments.go
```diff
@@ -230,16 +231,38 @@ func (s *DatastoreSegment) End() {
 	s.StartTime.end()
 }
 
+var (
+	hostsToReplace = map[string]struct{}{
```
newrelic-eheinlein, line 235: It seemed a little weird to me at first to use a map with empty values for this, but I can see how that would be the fastest and best way. My impulse was to use a Set, but that's a Java thing, not a Go thing. Still getting rid of my old habits.

purple4reina: Yeah, Go doesn't have Sets, but it does have maps with empty values. And for the record, I copied this from the v3 code anyway.

PR #222, opened 2020-08-19
Add all required transaction level attributes.

purple4reina:
NOTE: This PR requires #216 so merge that one first.

This pull request adds all transaction level attributes to the root span. It also sets the span status based on the response code.

There still needs to be more testing for the `txn.SetWebRequest`, `txn.SetWebRequestHTTP`, and `txn.SetWebResponse` methods, particularly for using nil transactions.


newrelic-eheinlein:
Well, I borked this up.  Sorry Rey! In order to save time, I redid this PR in #225 .

PR #229, opened 2020-09-03
Fix go-fmt issue

newrelic-eheinlein:
A white space issue was introduced directly into master accidentally. This fixes that issue.


joshuabenuck (approved):

newrelic-eheinlein:
Note: the one failing test is due to an issue that is resolved on the develop branch, so we are going to ignore that one failing test.

PR #230, opened 2020-09-11
Move SQL query parsing into separate goroutine

newrelic-eheinlein:
This PR is not ready for merge, but I wanted to make some notes on it. The goal here is to take the heavy lifting of Regex SQL parsing and move it into a separate goroutine. Things left to do:

- [ ] The main issue is that right now the datastore segment isn't actually getting updated with the **operation/collection**. Perhaps I passed something by value instead of by reference? Those fields are getting parsed, and set on **a** segment according to my print statements it's just not getting set properly on the final segment.
- [ ] Do rigorous testing around  deadlocks/ensure data is correct if the segment is waiting when a harvest happens
- [ ] Check: is this acutally helpful in the `Prepare` case? It should help when actually making a call to a db using `Exec` or `Query`, but if it is only using `Prepare` then it might actually slow that down because it doesn't have a database call to wait for as well.
- [ ] I left my `fmt.Printf` statements in. Maybe they'll be helpful, maybe I'm just lazy. Anyway, those should be removed
- [ ] Rebase and change the base branch to `develop`.

### DIFF
--- a/v3/newrelic/segments.go
+++ b/v3/newrelic/segments.go
@@ -5,6 +5,7 @@ package newrelic
 
 import (
 	"net/http"
+	"sync"
 )
 
 // SegmentStartTime is created by Transaction.StartSegmentNow and marks the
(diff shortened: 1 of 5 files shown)


Miriam-R:
Is this still relevant in light of of #232 https://github.com/newrelic/go-agent/pull/232

RichVanderwal:
In the words of @nr-swilloughby, we'd "have to get a lot of things just right" for this to be trustworthy. I agree! Let's shelve this idea, knowing that the improvements in PR #232 its related issue, #301 , would help out without being disruptive to our existing customers.

PR #232, opened 2020-09-17
Optimize ParseQuery function

pyohannes:
In some instances, usages of `regex` can be very slow. This can cause considerable overhead in the `ParseQuery` function, which can use up to 6 evaluations of regular expressions.

This PR replaces the regular expressions with plain string operations. The main part consists in introducing a naive SQL tokenizer, which can extract SQL tokens from a string.

All unit tests pass, and the new algorithm is about 3 times faster, some benchmarks show it to be almost 10 times faster:

```
pkg: github.com/newrelic/go-agent/v3/newrelic/sqlparse
BenchmarkSqlParsing-4                         30             35010 ns/op             296 B/op         36 allocs/op
BenchmarkSqlParsingWithRegex-4                 3            338403 ns/op            1946 B/op         14 allocs/op
```

### DIFF
--- a/v3/newrelic/sqlparse/sqlparse.go
+++ b/v3/newrelic/sqlparse/sqlparse.go
@@ -4,50 +4,80 @@
 package sqlparse
 
 import (
+	"bytes"
 	"fmt"
-	"regexp"
 	"strings"
 	"unicode"
 
 	newrelic "github.com/newrelic/go-agent/v3/newrelic"
 )
 
+var (
+	operations = map[string]string{
+		"select":   "from",
+		"delete":   "from",
+		"insert":   "into",
+		"update":   "",
+		"call":     "",
+		"create":   "",
+		"drop":     "",
+		"show":     "",
+		"set":      "",
+		"exec":     "",
+		"execute":  "",
+		"alter":    "",
+		"commit":   "",
+		"rollback": "",
+	}
+	updateModifiers = map[string]interface{}{
+		"low_priority": nil,
+		"ignore":       nil,
+		"or":           nil,
+		"rollback":     nil,
+		"abort":        nil,
+		"replace":      nil,
+		"fail":         nil,
+		"only":         nil,
+	}
+)
+
+// Extracts the table name from the given string.
 func extractTable(s string) string {
-	s = extractTableRegex.ReplaceAllString(s, "")
 	if idx := strings.Index(s, "."); idx > 0 {
 		s = s[idx+1:]
 	}
-	return s
+
+	var buffer bytes.Buffer
+
+	for _, c := range s {
+		if unicode.IsSpace(c) || strings.ContainsRune("`'(){}[]\"", c) {
+			continue
+		} else {
+			buffer.WriteRune(c)
+		}
+	}
+
+	return buffer.String()
 }
 
-var (
-	basicTable        = `[^)(\]\[\}\{\s,;]+`
-	enclosedTable     = `[\[\(\{]` + `\s*` + basicTable + `\s*` + `[\]\)\}]`
-	tablePattern      = `(` + `\s+` + basicTable + `|` + `\s*` + enclosedTable + `)`
-	extractTableRegex = regexp.MustCompile(`[\s` + "`" + `"'\(\)\{\}\[\]]*`)
-	updateRegex       = regexp.MustCompile(`(?is)^update(?:\s+(?:low_priority|ignore|or|rollback|abort|replace|fail|only))*` + tablePattern)
-	sqlOperations     = map[string]*regexp.Regexp{
-		"select":   regexp.MustCompile(tablePattern),
-		"delete":   regexp.MustCompile(tablePattern),
-		"insert":   regexp.MustCompile(tablePattern),
-		"update":   updateRegex,
-		"call":     nil,
-		"create":   nil,
-		"drop":     nil,
-		"show":     nil,
-		"set":      nil,
-		"exec":     nil,
-		"execute":  nil,
-		"alter":    nil,
-		"commit":   nil,
-		"rollback": nil,
+// Returns a new string with trailing comments removed.
+func cutComment(query string) string {
+	query = skipSpace(query)
+	for i, c := range query {
+		if strings.HasPrefix(query[i:], "/*") || strings.HasPrefix(query[i:], "--") || c == ';' || c == '#' {
+			query = fmt.Sprintf("%s%s", query[:i], skipComment(query[i:]))
+			return cutComment(query)
+		}
 	}
-	firstWordRegex   = regexp.MustCompile(`^\w+`)
-	cCommentRegex    = regexp.MustCompile(`(?is)/\*.*?\*/`)
-	lineCommentRegex = regexp.MustCompile(`(?im)(?:--|#).*?$`)
-	sqlPrefixRegex   = regexp.MustCompile(`^[\s;]*`)
-)
+	return query
+}
 
+type sqlQuery struct {
+	q string
+	p uint64
+}
+
+// Returns a string slice of query without trailing spaces.
 func skipSpace(query string) string {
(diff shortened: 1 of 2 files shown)


v3/newrelic/sqlparse/sqlparse.go
```diff
@@ -4,47 +4,168 @@
...
+type sqlTokenizer struct {
+	query string
+}
+
+// Returns the first word and the remainder of the given string (with trailing
+// comments removed).
+//
+// A word is either sequence of letters and digits, or a single special
+// character.
+func (q *sqlTokenizer) nextWord() string {
```
rvanderwal-newrelic, line 116: I haven't thought this through all the way, but is there a reason why we get tokens "manually" here instead of using strings.Fields?

pyohannes: Because SQL is nasty. This, for example, is a valid SQL query:
```sql
select*from table_name
```
And `strings.Fields` only splits at spaces.

v3/newrelic/sqlparse/sqlparse.go
```diff
@@ -4,47 +4,168 @@
...
-		"create":   nil,
-		"drop":     nil,
-		"show":     nil,
-		"set":      nil,
-		"exec":     nil,
-		"execute":  nil,
-		"alter":    nil,
-		"commit":   nil,
-		"rollback": nil,
+// Returns a string slice of query without trailing spaces.
```
joshuabenuck, line 63: This looks like it is skipping leading spaces, not trailing spaces.

v3/newrelic/sqlparse/sqlparse.go
```diff
@@ -4,47 +4,168 @@
...
 	}
-	firstWordRegex   = regexp.MustCompile(`^\w+`)
-	cCommentRegex    = regexp.MustCompile(`(?is)/\*.*?\*/`)
-	lineCommentRegex = regexp.MustCompile(`(?im)(?:--|#).*?$`)
-	sqlPrefixRegex   = regexp.MustCompile(`^[\s;]*`)
-)
+	return query
+}
+
+// Returns a string slice of query with trailing comments removed.
```
joshuabenuck, line 73: This looks like it is skipping leading comments, not trailing comments.

v3/newrelic/sqlparse/sqlparse.go
```diff
@@ -55,15 +176,32 @@ var (
 // Ability to correctly parse queries for other SQL databases is not
 // guaranteed.
 func ParseQuery(segment *newrelic.DatastoreSegment, query string) {
-	s := cCommentRegex.ReplaceAllString(query, "")
-	s = lineCommentRegex.ReplaceAllString(s, "")
-	s = sqlPrefixRegex.ReplaceAllString(s, "")
-	op := strings.ToLower(firstWordRegex.FindString(s))
-	if rg, ok := sqlOperations[op]; ok {
+	sql := sqlTokenizer{query: query}
+	op := strings.ToLower(removeAllComments(sql.nextWord()))
```
joshuabenuck, line 180: Why are we calling `removeAllComments` when `nextWord` already skips leading and trailing comments? Won't this just result in unnecessary scans of the word?

v3/newrelic/sqlparse/sqlparse.go
pyohannes: Fixed.

v3/newrelic/sqlparse/sqlparse.go
pyohannes: You're right, that's unnecessary. I removed the call.

v3/newrelic/sqlparse/sqlparse.go
pyohannes: Fixed.

v3/newrelic/sqlparse/sqlparse.go
```diff
@@ -55,15 +176,32 @@ var (
...
-				segment.Collection = extractTable(m[1])
+		if tablePrefix != "" {
+			for {
+				word := sql.nextWord()
+				if word == "" {
+					break
+				}
+				if strings.ToLower(word) == tablePrefix {
+					table := sql.nextToken()
+					segment.Collection = extractTable(removeAllComments(table))
```
joshuabenuck, line 191: Won't the call to `nextToken` result in a token free of comments?

joshuabenuck: .... Answering my own question. I think a comment could be between the `[` and the `]`.

v3/newrelic/sqlparse/sqlparse.go
```diff
@@ -55,15 +176,32 @@ var (
...
+				}
```
joshuabenuck, line 192: The `tablePrefix` clause and the `op == "update"` clause are mutually exclusive. Would there be value in returning early here to make that more apparent?

pyohannes: I chaned the `if`s.

v3/newrelic/sqlparse/sqlparse_test.go
```diff
@@ -192,3 +192,91 @@ func TestExtractTable(t *testing.T) {
...
+	}
+}
+
+func TestRemoveAllComments(t *testing.T) {
+	for input, expected := range map[string]string{
+		"query -- comment":                                  "query  ",
+		"query -- comment\nother":                           "query  other",
+		"query /* comment */":                               "query  ",
+		"query /* comment */ other":                         "query  other",
+		"query /* comment */ other /* other comment */ end": "query  other  end",
```
joshuabenuck, line 245: I think it would be good to have a test for leading and trailing comments in this set. Perhaps just add a comment at the beginning and the end of this test case?

```suggestion
		"/* leading */ query /* comment */ other /* other comment */ end": "query  other  end /* trailing */",
```

pyohannes: I added this test case. I had to add spaces at the end of the result string.

joshuabenuck (approved):

blakeroberts-wk:
Could the benchmark tests be added to the unit test file? It'd be nice to reproduce and verify the results

pyohannes:
@blakeroberts-wk 

>Could the benchmark tests be added to the unit test file? It'd be nice to reproduce and verify the results

I added the benchmark to the unit tests (see [here](https://github.com/newrelic/go-agent/pull/232/files#diff-b0d887b40d2d620cf00c292a58702d5dR285)). You'd need to back-port it to measure the unoptimized version of `ParseQuery`, as it is replaced in this revision.

v3/newrelic/sqlparse/sqlparse.go
```diff
@@ -4,47 +4,168 @@
...
-	s = extractTableRegex.ReplaceAllString(s, "")
 	if idx := strings.Index(s, "."); idx > 0 {
 		s = s[idx+1:]
 	}
-	return s
+
+	var buffer bytes.Buffer
+
+	for _, c := range s {
+		if unicode.IsSpace(c) || strings.ContainsRune("`'(){}[]\"", c) {
```
blakeroberts-wk, line 53:
```suggestion
		if unicode.IsSpace(c) || strings.ContainsRune("`'(){}[]\";", c) {
```

blakeroberts-wk: if you do `SELECT a,b FROM table;` you get the collection `table;`

pyohannes: Great catch, thanks for the fix. I also added a test for that case.

RichVanderwal:
We're waiting for Workiva to get back to us to see if this has helped them. See #workiva-424963-escalation for details.

Ak-x:
@pyohannes is this still a relevant PR, if so we can allocate time for our team to review.

The CLA still needs to be signed.

PR #236, opened 2020-09-28
WIP: NR Gin Integration support retrieving transaction from context

bernardoVale:
## Details

We needed a way to instrument middlewares that are part of the request chain but before Gin Router.  We could, of course, start and end the transaction before the router, but then we would be measuring the time incorrectly for all endpoints since there was some computation outside the endpoint transaction.

The use case is easy to understand, we have a middleware for path rewrite, this kind of feature needs to be executed before the router. Gin doesn't provide any primitives for "pre routing" middlewares, so if you need that, you have to create a regular http.Handler and attach it before the router.  Like this:

```go
type testParentMiddleware struct {
	app *newrelic.Application
	key string
	next http.Handler
}

func (t *testParentMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
	txn := t.app.StartTransaction("parent")
	sgm := txn.StartSegment("parent-segment")
	txn.SetWebRequestHTTP(r)
	// imagine some computation here
	r = r.WithContext(context.WithValue(r.Context(), t.key, txn))
	sgm.End()
	t.next.ServeHTTP(rw, r)
}

func main() {
  app := NewNewrelicApp()
  key := "newrelic"
  router := gin.Default()
  router.Use(nrgin.MiddlewareWithTransaction(app, key))
  handler := testParentMiddleware{
		app:  app.Application,
		key: key,
		next: router,
   }
  // attach handler to an http.Server{}
}
```

This is the best solution we found so far, start the transaction in the first middleware and adapt `nrgin` to conditionally retrieve the transaction from `http.Request` context.

Before writing tests and polishing docs, I want to receive some initial feedback so I can avoid wasting my time.

### DIFF
--- a/v3/integrations/nrgin/nrgin.go
+++ b/v3/integrations/nrgin/nrgin.go
@@ -123,6 +123,28 @@ func Middleware(app *newrelic.Application) gin.HandlerFunc {
 	return middleware(app, true, getTransactionNew)
 }
 
+// MiddlewareWithTransaction creates a Gin middleware that instruments
+// requests allowing the client to optionally start the Transaction before
+// this middleware. This functionally is particularly useful if you want
+// to instrument middleware that executes before Gin.
+//
+//
+//
+//  router := gin.Default()
+//  // Add the nrgin middleware before other middlewares or routes:
+//	router.Use(nrgin.MiddlewareWithTransaction(app, "context-key-associated-with-your-transaction"))
+//
+// Gin v1.5.0 introduced the gin.Context.FullPath method which allows for much
+// improved transaction naming.  This Middleware will use that
+// gin.Context.FullPath if available and fall back to the original
+// gin.Context.HandlerName if not.  If you are using Gin v1.5.0 and wish to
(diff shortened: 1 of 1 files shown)


rvanderwal-newrelic:
Thanks for the PR, @bernardoVale ! I'll be looking at this so we can address any concerns we have with this approach.

bernardoVale:
@rvanderwal-newrelic any updates? We're running 100% of production traffic with this code by the way :)

matheus-meneses:
Hi Guys, any update about this feature?

jodeev:
Thank you for your patience. I've heard interest in this from others as well. 

We'll have another look at this, we haven't forgotten it.

matheus-meneses:
Hello Guys, any updates about this PR?

nr-swilloughby:
Apologies for the delay on this. We're looking at it and should have an update for you soon.

Ak-x:
@matheus-meneses if this is still of interest are you able to resolve the merge conflict?

PR #249, opened 2020-11-25
Implement Amazon's IMDSv2

RichVanderwal:
This implements a newer method of authentication to AWS instance metadata endpoints. 

Under IMDSv1, the agent simply called to an endpoint and received JSON with information about the instance size, availability zone, and other facts.

This newer IMDSv2 implements a session protocol. The agent calls two endpoints: one to get a session token that's valid for a specified time, and then the original endpoint with the session token to get the instance information.

As of this writing, AWS normally accepts both IMDSv1 and IMDSv2. A customer can tell AWS to only allow IMDSv2, so we're fine if we simply perform IMDSv2 calls exclusively.

Tests for these changes will need to be higher-level integration tests, since they rely on AWS services. Let me know if you think otherwise!

## Links
[Amazon's IMDSv2 Documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html#configuring-instance-metadata-options)
[New Relic Java Agent Implementation](https://github.com/newrelic/newrelic-java-agent/blob/main/newrelic-agent/src/main/java/com/newrelic/agent/utilization/AWS.java)

## Details
One decision still in flight is what the timeout values should be for each call to the two endpoints. Since the original timeout was one second, I've split the difference and set the timeout for both calls to 500 milliseconds. These timeouts could be reduced, since the endpoints are reported to be very fast by the documentation above (< 10ms). Other New Relic language agents have other implementations, including a 100 millisecond timeout, and a back-off scheme using 15, 30, 60, 120, and 300 milliseconds.

### DIFF
--- a/v3/internal/utilization/aws.go
+++ b/v3/internal/utilization/aws.go
@@ -12,9 +12,11 @@ import (
 )
 
 const (
-	awsHostname     = "169.254.169.254"
-	awsEndpointPath = "/2016-09-02/dynamic/instance-identity/document"
-	awsEndpoint     = "http://" + awsHostname + awsEndpointPath
+	awsHostname          = "169.254.169.254"
+	awsEndpointPath      = "/2016-09-02/dynamic/instance-identity/document"
+	awsTokenEndpointPath = "/latest/api/token"
+	awsEndpoint          = "http://" + awsHostname + awsEndpointPath
+	awsTokenEndpoint     = "http://" + awsHostname + awsTokenEndpointPath
 )
 
 type aws struct {
@@ -44,6 +46,27 @@ func (e unexpectedAWSErr) Error() string {
 	return fmt.Sprintf("unexpected AWS error: %v", e.e)
 }
 
+// getAWSToken attempts to get the IMDSv2 token within the providerTimeout set
+// provider.go.
+func getAWSToken() (token string, err error) {
+	client := http.Client{
+		Timeout: providerTimeout,
+	}
+	request, err := http.NewRequest("PUT", awsTokenEndpoint, nil)
+	request.Header.Add("X-aws-ec2-metadata-token-ttl-seconds", "60")
+	response, err := client.Do(request)
+	if err != nil {
+		return "", err
+	}
+	defer response.Body.Close()
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return "", err
+	}
+
+	return string(body), nil
+}
+
 func getAWS(client *http.Client) (ret *aws, err error) {
 	// In some cases, 3rd party providers might block requests to metadata
 	// endpoints in such a way that causes a panic in the underlying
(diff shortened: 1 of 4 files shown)


v3/internal/utilization/aws.go
```diff
@@ -56,7 +77,18 @@ func getAWS(client *http.Client) (ret *aws, err error) {
 		}
 	}()
 
-	response, err := client.Get(awsEndpoint)
+	// AWS' IMDSv2 requires us to get a token before requesting metadata.
+	awsToken, err := getAWSToken(client)
+	if err != nil {
+		ret = nil
+		err = unexpectedAWSErr{e: fmt.Errorf("error contacting AWS IMDSv2 token endpoint, %v", err)}
```
krnowak, line 84: We should return here, if an error happened, right?
```suggestion
		return nil, unexpectedAWSErr{e: fmt.Errorf("error contacting AWS IMDSv2 token endpoint, %v", err)}
```

v3/internal/utilization/aws.go
```diff
@@ -12,9 +12,12 @@ import (
...
 const (
-	awsHostname     = "169.254.169.254"
-	awsEndpointPath = "/2016-09-02/dynamic/instance-identity/document"
-	awsEndpoint     = "http://" + awsHostname + awsEndpointPath
+	awsHostname          = "169.254.169.254"
+	awsEndpointPath      = "/2016-09-02/dynamic/instance-identity/document"
+	awsTokenEndpointPath = "/latest/api/token"
+	awsEndpoint          = "http://" + awsHostname + awsEndpointPath
+	awsTokenEndpoint     = "http://" + awsHostname + awsTokenEndpointPath
+	awsTokenTTL          = "60" // seconds this AWS utiliation session will last
```
krnowak, line 20:
```suggestion
	awsTokenTTL          = "60" // seconds this AWS utilization session will last
```

v3/internal/utilization/aws.go
krnowak: Or, maybe you could create a request to `awsEndpoint` before calling `getAWSToken` and add the token to the request's headers only if `err != nil`. This would make IMDSv2 optional in case someone wants to stick with IMDSv1 (if that is even possible).

krnowak (commented):
I know it's not ready for review, but just found it in the review queue in email, so just had a quick look.

v3/internal/utilization/aws.go
RichVanderwal: Thanks for noticing this! After thinking harder about it, we shouldn't throw an unExpectedAWSErr at all, we're expecting a timeout unless they're in AWS. Fixing this and the test.

A more thorough implementation would raise errors on other HTTP errors, but I'm deferring that work in the interest of time.

RichVanderwal: Regarding the extra request to awsEndpoint, I'd rather avoid making three calls to AWS and increasing the overhead by another 0.5 second on startup. As of now, IMDSv2 endpoints are always available -- the only control the customer has is to turn IMDSv1 off to help their security posture.

v3/internal/utilization/aws.go
RichVanderwal: Fixed.

MrAlias (approved):

PR #281, opened 2021-03-18
269 licenses

nr-swilloughby:
## Details
Every place we have a `go.mod` file, the pkg.go.dev site requires a `LICENSE.txt` file to also be present. This PR adds `LICENSE.txt` files as needed to satisfy this requirement.


nr-swilloughby:
Need to correct and resubmit this request. Withdrawing this one first.

Issue #287, opened 2021-03-24
"nil dereference" when using DAX


odannyc:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

When using DAX to query a dynamodb table and instrumenting the request with newrelic, I get a nil reference panic on this line:
https://github.com/newrelic/go-agent/blob/master/_integrations/nrawssdk/v1/nrawssdk.go#L28

```golang
internal.EndSegment(ctx, req.HTTPResponse.Header)
```

I stepped/debugged through a request and it looks like `HTTPResponse` is nil for DAX requests.

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your New Relic account or API keys! )

## Steps to Reproduce

1. Create a dynamodb request
2. Instrument the handlers with the newrelic helper `nrawssdk.InstrumentHandlers(&req.Handlers)`
3. Send the request with DAX

[NOTE]: # ( Please be as specific as possible. )

## Expected Behavior

Request sends without error and I see the trace in New Relic

[NOTE]: # ( Tell us what you expected to happen. )

## NR Diag results

panic!

[NOTE]: # ( Provide any other relevant log data. )

## Your Environment

- github.com/newrelic/go-agent/v3/integrations/nrawssdk-v1 v1.0.0

[TIP]:  # ( Include as many relevant details about your environment as possible including the running version of New Relic software and any relevant configurations. )

## Reproduction case

[TIP]: # ( Link a sample application that demonstrates the issue. )

## Additional context

[TIP]:  # ( Add any other context about the problem here. )

odannyc:
@RichVanderwal 

Please let me know if there is any additional information you need.

Thank you.

RichVanderwal:
Hi @odannyc , thanks for your patience here! You're on the short list of support issues for us to get back to, and we'll be taking a look at this soon.

RichVanderwal:
After some review, we've decided to check for a nil value here and skip adding the AWS Request ID span attribute if there's no response. PR forthcoming!

odannyc:
Hi @RichVanderwal ,

I have a PR open for this issue here:

https://github.com/newrelic/go-agent/pull/291

RichVanderwal:
Hello @odannyc ! We've included your PR  in the maintenance release of the Go Agent today. Let us know if you have any feedback, questions, or comments!

PR #291, opened 2021-03-30
Fixes issue with nil HttpResponse causing panics

odannyc:
## Links

https://github.com/newrelic/go-agent/issues/287
https://github.com/aws/aws-dax-go/issues/32

## Details

During the `endSegment` handler we need to check if `HTTPResponse` is populated. This prevents us from nil dereference panicking when requests to AWS services don't use HTTP such as DAX


nr-swilloughby (approved):
Thank you for submitting this code change.

PR #335, opened 2021-06-30
implemented JSON distributed trace converter

nr-swilloughby:
## Summary
When making calls to a transaction's `AcceptDistributedTraceHeaders()` you are required to pass a valid `http.Header` value holding the distributed trace headers you want the current transaction to use. However, sometimes this data arrives in a text form such as a JSON string.

This requires the programmer to perform multiple steps to build up the `http.Header` value, which can be prone to errors. I propose a convenience function that constructs this value in one step.

## Changes
This PR adds two functions. `DistributedTraceHeadersFromJson()` converts a JSON representation of the trace headers to an `http.Header` object.

`AcceptDistributedTraceHeadersFromJson()` does the above plus calls `AcceptDistributedTraceHeaders()` on the result for you.

## Details
Given a JSON string such as:
```
{
   "traceparent": "frob",
   "tracestate": "blorfl",
   "newrelic": "xyzzy"
}
```
it will output the `http.Header` value with three header keys (`traceparent`, `tracestate`, and `newrelic`) with corresponding values as provided in the JSON string.

It will also appropriately handle the case where multiple values are provided in the JSON string for a single key.

An error will be returned if it was unable to parse the JSON string or ran into other difficulties.

### DIFF
--- a/v3/go.mod
+++ b/v3/go.mod
@@ -3,6 +3,6 @@ module github.com/newrelic/go-agent/v3
 go 1.7
 
 require (
-	github.com/golang/protobuf v1.4.3
-	google.golang.org/grpc v1.39.0
+	github.com/golang/protobuf v1.3.3
+	google.golang.org/grpc v1.27.0
 )
--- a/v3/go.mod
+++ b/v3/go.mod
@@ -3,6 +3,6 @@ module github.com/newrelic/go-agent/v3
 go 1.7
 
 require (
-	github.com/golang/protobuf v1.3.3
-	google.golang.org/grpc v1.27.0
+	github.com/golang/protobuf v1.4.3
+	google.golang.org/grpc v1.39.0
 )
--- a/v3/newrelic/trace_context_test.go
+++ b/v3/newrelic/trace_context_test.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"reflect"
 	"strings"
 	"testing"
 
@@ -44,6 +45,47 @@ type TraceContextTestCase struct {
 		Span             *fieldExpect `json:"Span,omitempty"`
 		TransactionError *fieldExpect `json:"TransactionError,omitempty"`
 	} `json:"intrinsics"`
+}
+
+func TestJsonDTHeaders(t *testing.T) {
+	type testcase struct {
+		in  string
+		out http.Header
+		err bool
+	}
+
+	for i, test := range []testcase{
+		{"", http.Header{}, false},
+		{"{}", http.Header{}, false},
+		{" invalid ", http.Header{}, true},
+		{`"foo"`, http.Header{}, true},
+		{`{"foo": "bar"}`, map[string][]string{
+			"Foo": {"bar"},
+		}, false},
+		{`{
+			"foo": "bar",
+			"baz": "quux",
+			"multiple": [
+				"alpha",
+				"beta",
+				"gamma"
+			]
+		}`, map[string][]string{
+			"Foo":      {"bar"},
+			"Baz":      {"quux"},
+			"Multiple": {"alpha", "beta", "gamma"},
+		}, false},
+	} {
+		h, err := DistributedTraceHeadersFromJson(test.in)
+
+		if err != nil {
+			if !test.err {
+				t.Errorf("case %d: %v: error expected but not generated", i, test.in)
+			}
+		} else if !reflect.DeepEqual(test.out, h) {
+			t.Errorf("case %d, %v -> %v but expected %v", i, test.in, h, test.out)
+		}
+	}
 }
 
 func TestCrossAgentW3CTraceContext(t *testing.T) {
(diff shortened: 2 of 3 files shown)


RichVanderwal:
Looks like we're getting errors on the [1.15.x test run](https://github.com/newrelic/go-agent/pull/335/checks?check_run_id=2956702231): 

```
Error: transaction.go:274:1: comment on exported method Transaction.AcceptDistributedTraceHeadersFromJson should be of the form "AcceptDistributedTraceHeadersFromJson ..."
Error: transaction.go:281:25: method AcceptDistributedTraceHeadersFromJson should be AcceptDistributedTraceHeadersFromJSON
Error: transaction.go:290:1: comment on exported function DistributedTraceHeadersFromJson should be of the form "DistributedTraceHeadersFromJson ..."
Error: transaction.go:305:6: func DistributedTraceHeadersFromJson should be DistributedTraceHeadersFromJSON
Error: transaction.go:328:24: error strings should not be capitalized or end with punctuation or a newline
Error: transaction.go:333:22: error strings should not be capitalized or end with punctuation or a newline
Error: transaction.go:338:20: error strings should not be capitalized or end with punctuation or a newline
```

v3/go.mod
```diff
@@ -3,6 +3,6 @@ module github.com/newrelic/go-agent/v3
 go 1.7
 
 require (
-	github.com/golang/protobuf v1.3.3
-	google.golang.org/grpc v1.27.0
+	github.com/golang/protobuf v1.4.3
+	google.golang.org/grpc v1.39.0
```
RichVanderwal, line 7: If we're requiring new versions of these libraries, will this affect our customers in any way?

v3/newrelic/trace_context_test.go
```diff
@@ -46,6 +47,47 @@ type TraceContextTestCase struct {
...
+		out http.Header
+		err bool
+	}
+
+	for i, test := range []testcase{
+		{"", http.Header{}, false},
+		{"{}", http.Header{}, false},
+		{" invalid ", http.Header{}, true},
+		{`"foo"`, http.Header{}, true},
+		{`{"foo": "bar"}`, map[string][]string{
```
RichVanderwal, line 62: Why is this `map[string][]string` rather than just another http.Header? I've changed them back to http.Header and all seems to work fine.

v3/newrelic/trace_context_test.go
```diff
@@ -46,6 +47,47 @@ type TraceContextTestCase struct {
...
+		}, false},
+		{`{
+			"foo": "bar",
+			"baz": "quux",
+			"multiple": [
+				"alpha",
+				"beta",
+				"gamma"
+			]
+		}`, map[string][]string{
```
RichVanderwal, line 73: Same here with `http.Header` vs. `map[string][]string`

v3/go.mod
nr-swilloughby: I think we can move the version back. That was automatically bumped by my go runtime, not because I purposefully upgraded it.

v3/newrelic/trace_context_test.go
nr-swilloughby: http.Header is more correct to use, even though they're (currently) equivalent.

v3/newrelic/transaction.go
```diff
@@ -269,6 +271,76 @@ func (txn *Transaction) AcceptDistributedTraceHeaders(t TransportType, hdrs http
...
+// DistributedTraceHeadersFromJson() takes a set of distributed trace headers as a JSON-encoded string
+// and emits a http.Header value suitable for passing on to the
+// txn.AcceptDistributedTraceHeaders() function.
+//
+// For example, given the input string
+//   `{"traceparent": "frob", "tracestate": "blorfl", "newrelic": "xyzzy"}`
+// This will emit an http.Header value with headers "traceparent", "tracestate", and "newrelic".
+//
+// This is a convenience function provided for cases where you receive the trace header data
+// already as a JSON string and want to avoid manually converting that to an http.Header.
```
RichVanderwal, line 300: I wonder if we want to add that this is likely when working with trace headers with different language agents? Maybe that would be too wordy. I just want to make sure somebody like Jo Ann would see this and use it.

v3/newrelic/transaction.go
```diff
@@ -269,6 +271,76 @@ func (txn *Transaction) AcceptDistributedTraceHeaders(t TransportType, hdrs http
...
+//
+// The JSON string must be a single object whose values may be strings or arrays of strings.
+// These are translated directly to http headers with singleton or multiple values.
+//
+func DistributedTraceHeadersFromJSON(jsondata string) (hdrs http.Header, err error) {
+	var raw interface{}
+	hdrs = http.Header{}
+	if jsondata == "" {
```
RichVanderwal, line 308: Should we check for `nil` here, as well as an empty string? I'm still wrapping my head around what errors are likely to be made when calling this function.

nr-swilloughby: A string value can't be nil.

v3/newrelic/transaction.go
nr-swilloughby: Not a bad idea. I don't mind leaning to more verbosity when it comes to documentation. You never know when that extra sentence or paragraph was just what someone was looking for.

RichVanderwal (approved):
👍

Issue #331, opened 2021-06-24
nrgrpc integration is now reporting non-errors as errors in the UI


odannyc:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your New Relic account or API keys! )

The latest version of the newrelic grpc integration package is reporting all grpc codes as errors. I believe the issue is here:

https://github.com/newrelic/go-agent/blob/master/v3/integrations/nrgrpc/nrgrpc_server.go#L85-L87

I think we should do filtering on grpc codes that are actual errors. Here is an example of what filtering another library does:

https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/logging/zap/options.go#L104-L144

## Steps to Reproduce

[NOTE]: # ( Please be as specific as possible. )

- return a NotFound error code from the grpc server

## Expected Behavior

[NOTE]: # ( Tell us what you expected to happen. )

- Does not show up in newrelic as an "error"

## NR Diag results

[NOTE]: # ( Provide any other relevant log data. )

## Your Environment

[TIP]:  # ( Include as many relevant details about your environment as possible including the running version of New Relic software and any relevant configurations. )

## Reproduction case

[TIP]: # ( Link a sample application that demonstrates the issue. )

## Additional context

[TIP]:  # ( Add any other context about the problem here. )

nr-swilloughby:
Thanks for raising this issue. We're looking into the best way to provide the greater granularity you are requesting so that instead of reporting any non-`OK` status as an error, we can selectively report them as errors, warnings, info, etc.

RichVanderwal:
Releasing this very soon!

RichVanderwal:
Hi @odannyc , we released the 3.14.0 version of the Go Agent, which has a configurable error handler for gRPC. 

We'd love it if you could give it a spin! [Here are the docs](https://pkg.go.dev/github.com/newrelic/go-agent/v3/integrations/nrgrpc#hdr-Server) for this new error handling.

PR #338, opened 2021-07-02
Enhanced handling of grpc errors (issue #331)

nr-swilloughby:
Amends the error handling capability of the nrgrpc integration to differentiate between grpc statuses which should be reported as errors with those to be reported as warnings or informational messages. 

## Details
Rather than calling `NoticeError()` on any non-`OK` status, we call `NoticeError()` only on ones we think merit that designation. Others are reported by adding attributes to the transaction which describe the grpc status codes returned from the external service.

If the built-in default actions are not sufficient or don't match the application's needs, they may be changed by the user by changing the designation of any specific grpc error, or even by supplying a custom handler function to perform any action desired when the corresponding grpc error is seen.

### DIFF
--- a/v3/integrations/nrgrpc/example/server/server.go
+++ b/v3/integrations/nrgrpc/example/server/server.go
@@ -14,8 +14,6 @@ import (
 	sampleapp "github.com/newrelic/go-agent/v3/integrations/nrgrpc/example/sampleapp"
 	newrelic "github.com/newrelic/go-agent/v3/newrelic"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 )
 
 // Server is a gRPC server.
@@ -30,8 +28,7 @@ func processMessage(ctx context.Context, msg *sampleapp.Message) {
 // DoUnaryUnary is a unary request, unary response method.
 func (s *Server) DoUnaryUnary(ctx context.Context, msg *sampleapp.Message) (*sampleapp.Message, error) {
 	processMessage(ctx, msg)
-	//	return &sampleapp.Message{Text: "Hello from DoUnaryUnary"}, nil
-	return &sampleapp.Message{}, status.New(codes.DataLoss, "oooooops!").Err()
+	return &sampleapp.Message{Text: "Hello from DoUnaryUnary"}, nil
 }
 
 // DoUnaryStream is a unary request, stream response method.
--- a/v3/integrations/nrgrpc/example/server/server.go
+++ b/v3/integrations/nrgrpc/example/server/server.go
@@ -14,6 +14,8 @@ import (
 	sampleapp "github.com/newrelic/go-agent/v3/integrations/nrgrpc/example/sampleapp"
 	newrelic "github.com/newrelic/go-agent/v3/newrelic"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 // Server is a gRPC server.
@@ -28,7 +30,8 @@ func processMessage(ctx context.Context, msg *sampleapp.Message) {
 // DoUnaryUnary is a unary request, unary response method.
 func (s *Server) DoUnaryUnary(ctx context.Context, msg *sampleapp.Message) (*sampleapp.Message, error) {
 	processMessage(ctx, msg)
-	return &sampleapp.Message{Text: "Hello from DoUnaryUnary"}, nil
+	//	return &sampleapp.Message{Text: "Hello from DoUnaryUnary"}, nil
+	return &sampleapp.Message{}, status.New(codes.DataLoss, "oooooops!").Err()
 }
 
 // DoUnaryStream is a unary request, stream response method.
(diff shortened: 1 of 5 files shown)


v3/integrations/nrgrpc/nrgrpc_server.go
```diff
@@ -42,6 +72,167 @@ func startTransaction(ctx context.Context, app *newrelic.Application, fullMethod
...
+}
+
+//
+// `Configure()` takes a list of `WithStatusHandler()` options and sets them
+// as the new default handlers for the specified gRPC status codes, in the same
+// way as if `WithStatusHandler()` were given to the `StreamServiceInterceptor()`
+// or `UnaryServiceInterceptor()` functions (q.v.); however, in this case the new handlers
+// become the default for any subsequent interceptors created by the above functions.
+//
+func Configure(options ...handlerOption) {
```
RichVanderwal, line 159: Do you think `Configure()` might be too general a name for this? There might be other things to configure in the future.

Oh, but I suppose we could have other "With..." options that have  nothing to do with StatusHandlers, and then the Configure method name would still make sense. I'm a little fuzzy on what this would look like.

nr-swilloughby: I was pondering that too. Since it's in the nrgrpc package, it would be just configuring things within the package, but the idea was that we'd add other "With..." options for other package configuration needs, so we don't have multiple calls to make to customize the package in various ways.

nr-swilloughby:
I had a thought about the spurious error reporting that's being triggered elsewhere being related to our setting the transaction's response code. We may decide not to call `NoticeError()` ourselves, but elsewhere in the framework the web return code may trigger an error report anyway. I just pushed a commit that always sets the transaction's response code to `codes.OK` unless we're in an error handler to see if that gives us the desired behavior. If that works, we should only see errors when we asked for a particular grpc code to mean "error" and not for "info" or "warning".

v3/integrations/nrgrpc/example/server/server.go
```diff
@@ -28,7 +30,8 @@ func processMessage(ctx context.Context, msg *sampleapp.Message) {
 // DoUnaryUnary is a unary request, unary response method.
 func (s *Server) DoUnaryUnary(ctx context.Context, msg *sampleapp.Message) (*sampleapp.Message, error) {
 	processMessage(ctx, msg)
-	return &sampleapp.Message{Text: "Hello from DoUnaryUnary"}, nil
+	//	return &sampleapp.Message{Text: "Hello from DoUnaryUnary"}, nil
+	return &sampleapp.Message{}, status.New(codes.DataLoss, "oooooops!").Err()
```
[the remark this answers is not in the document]
nr-swilloughby: Yes, good catch.

v3/integrations/nrgrpc/nrgrpc_server.go
```diff
@@ -183,8 +183,14 @@ func OKInterceptorStatusHandler(ctx context.Context, txn *newrelic.Transaction,
 // contextual information gleaned from the error value received from the RPC call.
 //
 func ErrorInterceptorStatusHandler(ctx context.Context, txn *newrelic.Transaction, s *status.Status) {
-	txn.SetWebResponse(nil).WriteHeader(int(s.Code()))
-	txn.NoticeError(s.Err())
+	txn.SetWebResponse(nil).WriteHeader(int(codes.OK))
+	txn.NoticeError(&newrelic.Error{
+		Message: s.Message(),
+		Class:   "gRPC Error: " + s.Code().String(),
```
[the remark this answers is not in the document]
nr-swilloughby: It would be more accurate to call it "gRPC Status Code", actually.

v3/integrations/nrgrpc/nrgrpc_server.go
```diff
@@ -42,6 +73,173 @@ func startTransaction(ctx context.Context, app *newrelic.Application, fullMethod
...
+	codes.ResourceExhausted:  WarningInterceptorStatusHandler,
+	codes.FailedPrecondition: WarningInterceptorStatusHandler,
+	codes.Aborted:            WarningInterceptorStatusHandler,
+	codes.OutOfRange:         WarningInterceptorStatusHandler,
+	codes.Unimplemented:      ErrorInterceptorStatusHandler,
+	codes.Internal:           ErrorInterceptorStatusHandler,
+	codes.Unavailable:        WarningInterceptorStatusHandler,
+	codes.DataLoss:           ErrorInterceptorStatusHandler,
+	codes.Unauthenticated:    InfoInterceptorStatusHandler,
+}
```
RichVanderwal, line 112: Looks like we have tests for `ErrorInterceptorStatusHandler` below, using `codes.DataLoss`. Would it make sense to include tests for the other handlers, as well?

v3/integrations/nrgrpc/nrgrpc_server.go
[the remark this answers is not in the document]
RichVanderwal: Oh, yeah. That's much better.

v3/integrations/nrgrpc/nrgrpc_server.go
RichVanderwal: Deferring for later, @nr-swilloughby 's putting something on the Go Engineering Board so it doesn't get forgotten.

RichVanderwal (approved):

PR #360, opened 2021-08-05
Added support for API Gateway V2 HTTP requests and responses

tiaanl:
## Links

https://github.com/aws/aws-lambda-go/blob/master/events/apigw.go#L51

## Details

Detech when API Gateway V2 HTTP requests/responses are returned from the lambda handler and do the same reporting as the normal API Gateway proxy requests/responses.

Bumped the version of `github.com/aws/aws-lambda-go` to 1.26.0, which is the latest at the time of this PR.

### DIFF
--- a/v3/integrations/nrlambda/events.go
+++ b/v3/integrations/nrlambda/events.go
@@ -54,6 +54,10 @@ func eventWebRequest(event interface{}) *newrelic.WebRequest {
 		request.Method = r.HTTPMethod
 		path = r.Path
 		headers = r.Headers
+	case events.APIGatewayV2HTTPRequest:
+		request.Method = r.RequestContext.HTTP.Method
+		path = r.RawPath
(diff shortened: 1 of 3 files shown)


nr-swilloughby:
@tiaanl, 
Thank you for your PR. I just have one question. In your submission, the `go.mod` file bumps the `aws-lambda-go` version from v1.11.0 to v1.26.0. Is there any reason this specific version is needed, or was it just the version that Go happened to load for you, and a lower version number would still work well. Just checking so we make sure we support the broadest range of products our customers may want to use.

tiaanl:
@nr-swilloughby 1.26 is not strictly required, but only version 1.16, so i added a commit to change to that version.

Ak-x:
@tiaanl are you able to sign the CLA? That will allow us to proceed.  We can re-open once that is signed.

Issue #397, opened 2021-10-14
Incorrect table name being parsed from query when using nrpq package


tsmith-rv:
When using the `nrpq` package to instrument my database calls with the `nrpostgres` driver, the incorrect table name is being parsed from my query

## Description

When I instrument my queries using the `nrpostgres` driver, if I have a query that uses `from` outside of the normal context of specifying a table name the data in newrelic shows whatever keyword / token comes right after the `from` as the table name.

## Steps to Reproduce

Instrument a query using the `nrpostgres` driver and use a query that uses the `from` keyword in the context of a function like `substring` or `trim`.

For example:

For `SELECT substring('Thomas' from 2 for 3) as b FROM data`, the table name will show up as `2`.

For `SELECT trim(both ' ' from to_char(id, '99999')) as b FROM data`, the table name will show up as `to_char`.

## Expected Behavior

Using the examples from above:

For `SELECT substring('Thomas' from 2 for 3) as b FROM data`, the table name should show up as `data`.

For `SELECT trim(both ' ' from to_char(id, '99999')) as b FROM data`, the table name should show up as `data`.

## NR Diag results

![Screen Shot 2021-10-14 at 9 24 53 AM](https://user-images.githubusercontent.com/39560403/137327577-73d8c19f-1390-4e35-9b5e-9a5e59be05f5.png)

## Your Environment

github.com/newrelic/go-agent/v3/integrations/nrpq v1.1.1

PR #399, opened 2021-10-19
(fix issue #397) incorrect table name reported

nr-swilloughby:
## Links
[Issue 397](https://github.com/newrelic/go-agent/issues/397)

## Details
Changes the regular expression used to recognize table name when doing simple SQL command parsing. Previously it looked for the first `FROM` but this was confused by subexpressions such as `SUBSTRING('foo' FROM x TO y)`. This changes the code to look for the last `FROM` instead.


RichVanderwal (approved):
Looks good. The tests we have should protect us from most changes in behavior that our customers might see from this change. Still, it's not extremely low risk, so we should be ready to patch this back the way it was if it starts to cause problems.

PR #362, opened 2021-08-18
Obfuscate License Key in Agent Logs

RichVanderwal:
Security-minded customers have noticed that the Go Agent logs the URL of the agent endpoint at New Relic. This URL includes the customer's license key. This PR removes most of the license key, but keeps the first and last two characters to facilitate troubleshooting.

This work is based on PR #354 submittted by @pete911 . Thanks so much for your contribution!

PR #415, opened 2021-12-02
Logger license

nr-swilloughby:
## Details

This PR adds a filter to messages logged through the go agent's `v3/internal/logger` facility, for all severity levels, to redact license key information from the message being logged. Since the log message may include the license key embedded in a URL string or other marshalled text, we do not try to limit the filter by parsing the strings. Rather, we simply replace the text "`license_key=`" followed by a hex digit sequence, such that the hex digits are removed.

### DIFF
--- a/v3/internal/logger/logger.go
+++ b/v3/internal/logger/logger.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"log"
 	"os"
+	"regexp"
 )
 
 // Logger matches newrelic.Logger to allow implementations to be passed to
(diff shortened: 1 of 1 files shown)


nr-swilloughby:
Is it worth considering removing the [previous license key redaction](https://github.com/newrelic/go-agent/pull/362) on the logic that this change will cover that use case as well, since this one sits at a more generic level (this redacts license keys from any logged message while that previous change was more specific)?

RichVanderwal (approved):
Steve and I reviewed this and I've been convinced that the "[redacted]" method of license key obfuscation is better than the previous method of obfuscation.

PR #418, opened 2021-12-03
Release 3.15.2

nr-swilloughby:
### Added
* Strings logged via the Go Agent's built-in logger will have strings of the form `license_key=`*hex-string* changed to `license_key=[redacted]` before they are output, regardless of severity level, where *hex-string* means a sequence of upper- or lower-case hexadecimal digits and dots ('.'). This incorporates [PR #415](https://github.com/newrelic/go-agent/pull/415).

### Support Statement
New Relic recommends that you upgrade the agent regularly to ensure that you’re getting the latest features and performance benefits. Additionally, older releases will no longer be supported when they reach end-of-life.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -231,3 +231,52 @@ jobs:
       env:
         DIRS: ${{ matrix.dirs }}
         EXTRATESTING: ${{ matrix.extratesting }}
+
+  go-agent-arm64:
+    # Run all unit tests on aarch64 emulator to ensure compatibility with AWS
+    # Graviton instances
+    runs-on: ubuntu-18.04
+    strategy:
+    # if one test fails, do not abort the rest
+      fail-fast: false
+      matrix:
+        include:
+          - go-version: 1.17.1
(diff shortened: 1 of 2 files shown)


v3/internal/logger/logger.go
```diff
@@ -72,8 +73,11 @@ func (f *logFile) fire(level, msg string, ctx map[string]interface{}) {
...
 	})
-	if nil == err {
-		f.l.Print(string(js))
+	if err == nil {
+		// scrub license keys from any portion of the log message
+		re := regexp.MustCompile(`license_key=[a-fA-F0-9.]+`)
+		sanitized := re.ReplaceAllLiteralString(string(js), "license_key=[redacted]")
+		f.l.Print(sanitized)
 	} else {
 		f.l.Printf("unable to marshal log entry: %v", err)
```
Ak-x, line 82: Would it be helpful to have a check outside as well on the off-chance that our key is printed

Ak-x (changes requested):

Ak-x (approved):

Issue #427, opened 2021-12-29
AWS retry requests fail with 403 SignatureDoesNotMatch when newrelic distributed tracing is enabled.


andriiberezhynskyi:
## Description

Every aws retry request fails with 403 SignatureDoesNotMatch when newrelic
distributed tracing is enabled.

## Steps to Reproduce

Here is a piece of code that reproduces the problem. In this example I inject
the request handler which forces first request attempt to always fail with
retry-able error (of course I don't do it in production, it's just to simulate
aws sdk request retries). Then the retried request fails with aws
SignatureDoesNotMatch error which indicates that newrelic agent has modified
retired request in such a why that it makes request's signature invalid.
How do I know it that the problem is caused by newrelic agent? If I remove the
line where I add newrelic txn to ctx then request is retried successfully.

```go
package main

import (
	"context"
	"log"
	"time"

	"github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/request"
	"github.com/aws/aws-sdk-go/aws/session"
	"github.com/aws/aws-sdk-go/service/s3"
	"github.com/aws/aws-sdk-go/service/s3/s3manager"
	"github.com/newrelic/go-agent/v3/integrations/nrawssdk-v1"
	"github.com/newrelic/go-agent/v3/newrelic"
)

func main() {
	s3bucket := "REDACTED"
	s3key := "REDACTED"
	nrLicenseKey := "REDACTED"
	nrAppName := "REDACTED"

	nrApp, err := newrelic.NewApplication(
		newrelic.ConfigAppName(nrAppName),
		newrelic.ConfigLicense(nrLicenseKey),

		func(config *newrelic.Config) {
			config.SpanEvents.Enabled = true
			config.DistributedTracer.Enabled = true
		},
	)
	if err != nil {
		log.Fatal(err)
	}

	if err := nrApp.WaitForConnection(5 * time.Second); nil != err {
		log.Fatal(err)
	}

	sess := session.Must(session.NewSessionWithOptions(session.Options{
		SharedConfigState: session.SharedConfigEnable,
	}))
	sess.Config.WithLogLevel(aws.LogDebugWithSigning | aws.LogDebugWithRequestErrors)

	nrawssdk.InstrumentHandlers(&sess.Handlers)

	// This forces the very first aws request to fail with retriable error.
	sess.Handlers.Send.SetBackNamed(request.NamedHandler{
[...]

PR #429, opened 2021-12-30
fix issue #427: move segment start from aws `send` step to `sign` step

molecule-man:
Add New Relic headers in a "Sign" aws handler rather than in a "Send" handler. This ensures the signature always includes the correct header values and therefore fixes issue #427


molecule-man:
closing this pr as it breaks other things https://github.com/newrelic/go-agent/issues/427#issuecomment-1003317086

Issue #432, opened 2022-01-07
unit tests for v3/integrations/nrawssdk-v1 are broken


rittneje:
The unit tests for nrawssdk-v1 are broken. This makes it very difficult to submit any fixes for that module.

One of the problems is that the code is expecting to find the AWS request id in in the _request_ header, but the test framework only sets it in the _response_ header. I believe the tests are correct here, meaning the code is actually broken. In particular, it was broken by #291.

Ak-x:
This should be resolved by the completion of: https://github.com/newrelic/go-agent/pull/433

Ak-x:
Closing this per #433

PR #433, opened 2022-01-07
do not include distributed trace headers in AWS SDK requests

### LINKED ISSUE #427: AWS retry requests fail with 403 SignatureDoesNotMatch when newrelic distributed tracing is enabled.

rittneje:
Fixes #427 and #432. Sending the distributed trace headers to AWS is rather pointless, and can cause the request signature to be corrupted on retries. Also, previously the AWS request id and response status code were not being properly captured.

Note: Due to breaking changes in the v3/internal package, the new release of nrawssdk-v1 will depend on the new release of v3. Not sure how this is normally dealt with.

### DIFF
--- a/v3/integrations/nrawssdk-v1/nrawssdk.go
+++ b/v3/integrations/nrawssdk-v1/nrawssdk.go
@@ -5,8 +5,6 @@
 package nrawssdk
 
 import (
-	"net/http"
-
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/newrelic/go-agent/v3/internal"
 	"github.com/newrelic/go-agent/v3/internal/awssupport"
@@ -26,14 +24,7 @@ func startSegment(req *request.Request) {
 }
 
 func endSegment(req *request.Request) {
-	ctx := req.HTTPRequest.Context()
-
-	hdr := http.Header{}
-	if req.HTTPRequest != nil {
-		hdr = req.HTTPRequest.Header
-	}
-
-	awssupport.EndSegment(ctx, hdr)
+	awssupport.EndSegment(req.HTTPRequest.Context(), req.HTTPResponse)
 }
 
 // InstrumentHandlers will add instrumentation to the given *request.Handlers.
--- a/v3/integrations/nrawssdk-v1/nrawssdk_test.go
+++ b/v3/integrations/nrawssdk-v1/nrawssdk_test.go
@@ -93,11 +93,12 @@ var (
 		},
 		UserAttributes: map[string]interface{}{},
 		AgentAttributes: map[string]interface{}{
-			"aws.operation": "Invoke",
-			"aws.region":    "us-west-2",
-			"aws.requestId": requestID,
-			"http.method":   "POST",
-			"http.url":      "https://lambda.us-west-2.amazonaws.com/2015-03-31/functions/non-existent-function/invocations",
+			"aws.operation":   "Invoke",
+			"aws.region":      "us-west-2",
+			"aws.requestId":   requestID,
+			"http.method":     "POST",
+			"http.url":        "https://lambda.us-west-2.amazonaws.com/2015-03-31/functions/non-existent-function/invocations",
+			"http.statusCode": 200,
 		},
 	}
 	externalSpanNoRequestID = internal.WantEvent{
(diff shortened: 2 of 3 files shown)


rittneje:
ping @nr-swilloughby

nr-swilloughby:
@rittneje thank you for reporting this issue and for suggesting a pull request to address it. We'll review the issue and work to resolve it ASAP.

RichVanderwal:
@rittneje Thanks for working on this solution that simply side-steps the whole issue. We still need to do some testing on this, and would welcome others' feedback on this solution!

RichVanderwal:
Tested with sample code, merging with develop.

RichVanderwal (approved):

PR #441, opened 2022-01-26
Handle attributes in nrpkgerrors

### LINKED ISSUE #409: v3/integrations/nrpkgerrors: support errorAttributer interface
currently nrpkgerrors supports `errorClasser` even though `pkg/errors` doesn't support `errorClasser`. I'm assuming that this is done so that any extensions on top of `pkg/errors` that add support for `errorClasser` will continue to work with nrpkgerrors.Wrap as well.

but unfortunately the same is not true for `errorAttributer` interface. nrpkgerrors wraps the error but if that error supports `errorAttributer` then that information would get lost.

## Desired Behaviour

if any error that supports `errorAttributer` interface is wrapped by `nrpkgerrors.Wrap` then the final wrapped error should also support `errorAttributer` interface.

## Possible Solution

add a call to a helper function below this line [here](https://github.com/newrelic/go-agent/blob/5be716d8cd050352243214f49b55042c652e2df7/v3/integrations/nrpkgerrors/nrpkgerrors.go#L86) that checks if the incoming error supports errorAttributer interface, if so then use the attributes given by that error.

iamemilio:
## Details

Fixes #409. nrpkgerrors.Wrap() now checks if the error it is passed has attributes, and if it does, copies them into the New Relic error it creates.

### DIFF
--- a/v3/integrations/nrpkgerrors/nrkpgerrors_test.go
+++ b/v3/integrations/nrpkgerrors/nrkpgerrors_test.go
@@ -31,6 +31,7 @@ func gamma(e error) error { return errors.WithStack(e) }
 
 func theta(e error) error { return errors.WithMessage(e, "theta") }
 
+func basicNRError(e error) newrelic.Error { return newrelic.Error{Message: e.Error()} }
 func withAttributes(e error) newrelic.Error {
 	return newrelic.Error{
 		Message: e.Error(),
@@ -54,6 +55,7 @@ func TestWrappedStackTrace(t *testing.T) {
 		{Error: alpha(theta(beta(basicError{}))), ExpectTopFrame: "beta"},
 		{Error: alpha(theta(beta(theta(basicError{})))), ExpectTopFrame: "beta"},
 		{Error: theta(basicError{}), ExpectTopFrame: ""},
+		{Error: basicNRError(basicError{}), ExpectTopFrame: ""},
 		{Error: withAttributes(basicError{}), ExpectTopFrame: "", ExpectAttributes: map[string]interface{}{"testAttribute": 1, "foo": 2}},
 	}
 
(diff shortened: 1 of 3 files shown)


v3/integrations/nrpkgerrors/nrkpgerrors_test.go
```diff
@@ -31,10 +31,22 @@ func gamma(e error) error { return errors.WithStack(e) }
...
 func TestWrappedStackTrace(t *testing.T) {
 	testcases := []struct {
-		Error          error
-		ExpectTopFrame string
+		Error            error
+		ExpectTopFrame   string
+		ExpectAttributes map[string]interface{}
 	}{
 		{Error: basicError{}, ExpectTopFrame: ""},
 		{Error: alpha(basicError{}), ExpectTopFrame: "alpha"},
```
RichVanderwal, line 52: Odd indentation with the next two lines -- are these line feeds or something? GitHub won't let me comment directly on them.

v3/newrelic/trace_observer_test.go
```diff
@@ -902,7 +902,7 @@ func TestTrObsOKSendBackoffNo(t *testing.T) {
 	}
 	// If the default backoff of 15 seconds is used, the second span will not
 	// be received in time.
-	if !s.DidSpansArrive(t, 2, 4*time.Second) {
+	if !s.DidSpansArrive(t, 2, 8*time.Second) {
```
RichVanderwal, line 905: I'd love to find a better solution to this. I've increased this timeout myself in a past PR.

RichVanderwal (approved):
Lookin' good!

v3/newrelic/trace_observer_test.go
Ak-x: Can we set a variable for the integer here?

nr-swilloughby (approved):

PR #442, opened 2022-01-26
use Host from http.Request

eightnoteight:
## Details

Currently we use `localhost` in the url to point to envoy, and add host header to the request so that envoy would route to the appropriate cluster.

Since the url always has localhost, our external segment is rather confusing. by using Host from http.Request the external segment should properly distinguish for different external services.

### DIFF
--- a/v3/newrelic/segments.go
+++ b/v3/newrelic/segments.go
@@ -286,6 +286,7 @@ func StartExternalSegment(txn *Transaction, request *http.Request) *ExternalSegm
 	s := &ExternalSegment{
 		StartTime: txn.StartSegmentNow(),
 		Request:   request,
+		Host:      request.Host,
 	}
 
 	if request != nil && request.Header != nil {


nr-swilloughby:
Thanks for the PR! We'll review this as soon as possible.

RichVanderwal:
We're still reviewing this PR, and have a comment. In the meantime, you know you can pull the value of the header into your external request segment in your code like this:

```go
es := newrelic.StartExternalSegment(txn, req)
es.Host = "www.example.com"
resp, err := http.DefaultClient.Do(req)
```

If this solves your problem, let us know! Otherwise we'll continue taking a closer look at this PR.

v3/newrelic/segments.go
```diff
@@ -286,6 +286,7 @@ func StartExternalSegment(txn *Transaction, request *http.Request) *ExternalSegm
 	s := &ExternalSegment{
 		StartTime: txn.StartSegmentNow(),
 		Request:   request,
+		Host:      request.Host,
```
RichVanderwal, line 289: With our experimentation, this doesn't actually reflect the value set in the "Host" header. Host just gets set to whatever was actually requested in the URL -- in your case, it would be "localhost." Perhaps you're expecting the request object to pull the header automatically?

nr-swilloughby: @eightnoteight Can you confirm if @RichVanderwal's suggestion adequately solves your need without changing the existing behavior of the SDK, or if it really is a better solution to make the proposed change? Thanks!

iamemilio:
@eightnoteight Please let us know if the answer we provided works for you

nr-swilloughby:
@eightnoteight Can you verify if @RichVanderwal's suggestion adequately solves your need as an alternative to this PR? If we don't hear otherwise, we'll assume that is the case and close the PR. Thanks.

nr-swilloughby:
Closing this PR since no response was received to @RichVanderwal's question about whether his input adequately resolved the issue.

PR #445, opened 2022-02-01
Test Dependency Pinning

iamemilio:
Creates a new directive in the test matrix, "pin", that allows you to pin the version of a library imported during that test to a specific version. This is best used in cases where that specific test needs a version pin of something. Pins that affect a number of tests are cleaner to implement in the bash script directly.

### DIFF
--- a/build-script.sh
+++ b/build-script.sh
@@ -22,14 +22,14 @@ pwd=$(pwd)
 # inputs
 # 1: repo pin; example: github.com/rewrelic/go-agent@v1.9.0
 pin_go_dependency() {
-  echo "Pinning: $1"
   if [[ ! -z "$1" ]]; then
+    echo "Pinning: $1"
     repo=$(echo "$1" | cut -d '@' -f1)
     pinTo=$(echo "$1" | cut -d '@' -f2)
     set +e
     go get -u "$repo" # this go get will fail to build
     set -e
-    cd $GOPATH/src/"$repo"
+    cd "$GOPATH"/src/"$repo"
     git checkout "$pinTo"
     cd -
   fi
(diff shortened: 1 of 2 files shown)


iamemilio:
Sorry, I was playing around with that a little, the commit history for this is a total mess. I wound up putting the -t back because we did in fact need it :)

iamemilio:
This is still not ready to go, I am playing around with the nats server dependency

iamemilio:
Ok, this is nice and readable now. I have given up on making the nats.io integrations stuff work for V2 right now, for the sake of my sanity. I think this is formally ready for review.

build-script.sh
```diff
@@ -44,31 +57,29 @@ for dir in $DIRS; do
...
-      set +e
-      go get -u golang.org/x/net/http2 # this go get will fail to build
-      set -e
-      cd $GOPATH/src/golang.org/x/net/http2
-      git checkout 7fd8e65b642006927f6cec5cb4241df7f98a2210
-      cd -
+      pin_go_dependency "google.golang.org/grpc@v1.31.0"
+      pin_go_dependency "golang.org/x/net/http2@7fd8e65b642006927f6cec5cb4241df7f98a2210"
 
+      # install protobuff once dependencies are resolved
```
RichVanderwal, line 69: s/protobuff/protobuf/

build-script.sh
```diff
@@ -44,31 +57,29 @@ for dir in $DIRS; do
...
-      git checkout v1.31.0
-      cd -
-
-      set +e
-      go get -u golang.org/x/net/http2 # this go get will fail to build
-      set -e
-      cd $GOPATH/src/golang.org/x/net/http2
-      git checkout 7fd8e65b642006927f6cec5cb4241df7f98a2210
-      cd -
+      pin_go_dependency "google.golang.org/grpc@v1.31.0"
```
RichVanderwal, line 66: Love the way you generalized the whole pinning mechanism to make this so much easier to read.

RichVanderwal (dismissed):
More green tests! 💚

build-script.sh
iamemilio: I basically made a simple bash version of `go get` 🤣

nr-swilloughby (approved):

PR #448, opened 2022-02-04
Repair V3 Tests

iamemilio:
Get v3 tests passing again.


v3/integrations/nrgraphgophers/nrgraphgophers_test.go
```diff
@@ -244,7 +244,6 @@ func TestQueryRequestError(t *testing.T) {
 	app.ExpectErrors(t, []internal.WantError{{
 		TxnName: "WebTransaction/Go/POST /",
 		Msg:     "graphql: something went wrong",
-		Klass:   "*errors.QueryError",
```
RichVanderwal, line 247: Why isn't this Klass returned in the error event anymore?

v3/newrelic/expect_implementation.go
```diff
@@ -11,9 +11,13 @@ import (
 	"github.com/newrelic/go-agent/v3/internal"
 )
 
-func validateStringField(v internal.Validator, fieldName, v1, v2 string) {
-	if v1 != v2 {
-		v.Error(fieldName, v1, v2)
+func validateStringField(v internal.Validator, fieldName, expect, actual string) {
```
RichVanderwal, line 14: Note that there's another `validateStringField` in internal/expect.go. It follows the old "v1 != v2" format.

v3/newrelic/expect_implementation.go
```diff
@@ -11,9 +11,13 @@ import (
...
+	// If an expected value is not set, we assume the user does not want to validate it
+	if expect == "" {
+		return
+	}
+	if expect != actual {
+		v.Error(fieldName, "incorrect: Expected:", expect, " Got:", actual)
```
RichVanderwal, line 20: Thanks for spelling this out for the person who reads the test results!

v3/newrelic/expect_implementation.go
```diff
@@ -442,13 +446,13 @@ func expectSlowQuery(t internal.Validator, slowQuery *slowQuery, want internal.W
...
 	}
 	uri, _ := slowQuery.txnEvent.Attrs.GetAgentValue(AttributeRequestURI, destTxnTrace)
-	validateStringField(t, "MetricName", slowQuery.DatastoreMetric, want.MetricName)
-	validateStringField(t, "Query", slowQuery.ParameterizedQuery, want.Query)
-	validateStringField(t, "TxnEvent.FinalName", slowQuery.txnEvent.FinalName, want.TxnName)
-	validateStringField(t, "request.uri", uri, want.TxnURL)
-	validateStringField(t, "DatabaseName", slowQuery.DatabaseName, want.DatabaseName)
-	validateStringField(t, "Host", slowQuery.Host, want.Host)
-	validateStringField(t, "PortPathOrID", slowQuery.PortPathOrID, want.PortPathOrID)
+	validateStringField(t, "MetricName", want.MetricName, slowQuery.DatastoreMetric)
```
RichVanderwal, line 449: Any particular reason we've swapped the order of the parameters? This makes them all the same order, I suppose?

iamemilio: Yeah, this puts them all in expected:actual order.

v3/newrelic/expect_implementation.go
iamemilio: oh, I should fix that too

v3/integrations/nrgraphgophers/nrgraphgophers_test.go
iamemilio: that is a good question..

iamemilio: Ah, I left myself a note in the commit message :smile:. The graphgophers library changed the error type that they return in a later version of this library. Because we test both a pinned version and the lates version, we need to be able to handle two different error types in this block of code, so I adjusted the logic to allow you to pass "" as the error klass which matches any error type. See the commit titled "Remove nrgraphgophers_test error type expectation" for this change.

nr-swilloughby (approved):

PR #455, opened 2022-02-11
Add support for OpenBSD

qbit:
This adds `PhysicalMemoryBytes` functions and tests for OpenBSD.


iamemilio (dismissed):
This looks good to me, its nearly identical to what we support for freebsd, @RichVanderwal  has verified that this works with OpenBSD in vagrant, go vet did not find any risky pointer conversions, and the tests look good. Thank you for the PR!

qbit:
Do I need to do anything else for this?

iamemilio (approved):

PR #478, opened 2022-04-06
Added noop operations for js/wasm build target

taufik-rama:
## Details

Added basic compilation support for `js/wasm` build target.

Currently, when we try to compile the project to be used for WASM, it will result in error
```
taufik-rama@dev:~/programming/go/src/github.com/newrelic/go-agent$ GOOS=js GOARCH=wasm go build
# github.com/newrelic/go-agent/internal/sysinfo
internal/sysinfo/usage_posix.go:20:17: undefined: syscall.Getrusage
internal/sysinfo/usage_posix.go:20:35: undefined: syscall.RUSAGE_SELF
```

This changes just add basic support so that at least `sysinfo.GetUsage()` & `sysinfo.PhysicalMemoryBytes()` can be used, even though currently it will just return "unsupported" error


iamemilio:
Thanks for bringing this to our attention and submitting a pull request! We will take a look and get back to you as soon as we can.

taufik-rama:
I've signed the CLA & got the confirmation page, don't really know if there's any issue since it's still on `Pending` status.

I did registered my Github email after the confirmation, since I just so happen to use different email on the commit, don't know if this relates to that

iamemilio:
Hi @taufik-rama we are going to close and then re-open this repo in order to allow us to re-run the tests :)

taufik-rama:
Sure!

CMIIW it seems that some tests were not passing for previous versions. Unfortunately I'm currently in the middle of some work and doesn't have some free time to take a look into those, though I'll be sure to come back to this

nr-swilloughby (approved):
The PR is approved; in addition it looks like some work is needed to clue in the automated checks that not all modules are compiled on every platform, hence the false positives about redefined functions.

PR #495, opened 2022-05-18
Enable DT by Default for the Go Agent

nr-swilloughby:
This code enables DT by default for the Go agent.


v3/examples/client/main.go
```diff
@@ -14,14 +14,14 @@ import (
 
 func doRequest(txn *newrelic.Transaction) error {
 	req, err := http.NewRequest("GET", "http://localhost:8000/segments", nil)
-	if nil != err {
+	if err != nil {
```
iamemilio, line 17: thanks for fixing this, these always bothered me

nr-swilloughby: As a matter of principle I've been fixing those as I encounter them. when doing other fixes.

RichVanderwal (approved):

PR #501, opened 2022-05-26
Harvest log events

iamemilio:
This is a work in progress pull request for adding logs in context support to the go agent, and the first framework that we support: zero log.


iamemilio:
I will squash commits after review

iamemilio:
Idk how I just did this, but I'm going to undo it

PR #524, opened 2022-06-27
Custom Events Limiter

### LINKED ISSUE #476: can not configure max samples stored for custom events
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

It is a standard protocol for other agents to allow users to configure the number of custom events stored with a configuration setting named: `"custom_insights_events.max_samples_stored": 1000,`. This option is not available in the Go agent config.

mirackara:
### Description
Added configuration option for limiting the amount of samples stored in a custom insights event.

### Usage
When creating a new application, use the following function to limit sample sizes

	app, err := newrelic.NewApplication(
		newrelic.ConfigAppName("Short Lived App"),
		newrelic.ConfigLicense(os.Getenv("NEW_RELIC_LICENSE_KEY")),
		newrelic.ConfigCustomInsightsEventsMaxSamplesStored(240), // Configuration for limiting samples stored
	)

### DIFF
--- a/app_run.go
+++ b/app_run.go
@@ -7,7 +7,7 @@ import (
 	"encoding/json"
 	"strings"
 	"time"
-	
+
 	"github.com/newrelic/go-agent/internal"
 )
 
--- a/app_run.go
+++ b/app_run.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"strings"
 	"time"
+	
 	"github.com/newrelic/go-agent/internal"
 )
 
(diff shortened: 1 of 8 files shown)


nr-swilloughby (dismissed):

iamemilio:
This seems to be causing a number of the legacy tests to fail, which it should not effect. I think we should get to the bottom of why that is happening before we merge anything.

iamemilio (changes requested):
This seems to be causing a number of the legacy tests to fail, which it should not effect. I think we should get to the bottom of why that is happening before we merge anything.

I noticed that some of your changes altered internal, rather than v3/internal. I suspect that a lot of your trouble is coming from that change.

nr-swilloughby:
Good catch!

mirackara:
Legacy tests are now good. Looks like some more of the v3 tests are failing however. Thanks for the comment!

RichVanderwal:
Thanks for looking at how the other agents are configured when you were naming this option!

iamemilio (dismissed):

iamemilio (approved):

PR #534, opened 2022-07-06
Test dependancies fixes

mirackara:

### DIFF
--- a/v3/examples/short-lived-process/main.go
+++ b/v3/examples/short-lived-process/main.go
@@ -25,6 +25,7 @@ func main() {
 	// Wait for the application to connect.
 	if err := app.WaitForConnection(5 * time.Second); nil != err {
 		fmt.Println(err)
+
 	}
 
 	// Do the tasks at hand.  Perhaps record them using transactions and/or
--- a/v3/integrations/nrgorilla/go.mod
+++ b/v3/integrations/nrgorilla/go.mod
@@ -7,5 +7,5 @@ go 1.12
 require (
 	// v1.7.0 is the earliest version of Gorilla using modules.
 	github.com/gorilla/mux v1.7.0
-	github.com/newrelic/go-agent/v3 v3.0.0
+	github.com/newrelic/go-agent/v3 v3.17.0
 )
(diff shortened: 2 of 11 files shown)


.github/workflows/ci.yaml
```diff
@@ -75,22 +75,16 @@ jobs:
...
             dirs: _integrations/nrhttprouter
           - go-version: 1.13.x
             dirs: _integrations/nrb3
           - go-version: 1.13.x
             dirs: _integrations/nrmongo
-          - go-version: 1.13.x
-            dirs: _integrations/nrmssql
 
           # v3 agent
           - go-version: 1.7.x
```
iamemilio, line 90: Lets remove tests for the v3 agent for go version 1.7 and 1.12 as well. They are well outside of the deprecation window.

mirackara: Should we remove just the 1.7 and 1.12 v3 tests or 1.7 through 1.12?

iamemilio: You know what, let's handle this in a separate issue. I don't want to get stuck log rolling this. I also want to make test for 1.16, 1.17, 1.18. We should also separate the log context tests out into their own test. Idk why thats bundled.

iamemilio:
This PR is good to merge. Thanks for fixing this Mirac, this will have a big impact on the team being able to deliver software!

iamemilio (dismissed):

v3/integrations/nrgin/go.mod
```diff
@@ -1,10 +1,10 @@
 module github.com/newrelic/go-agent/v3/integrations/nrgin
 
-// As of Dec 2019, the gin go.mod file uses 1.12:
+// As of Jul 2022, the gin go.mod file uses 1.18:
 // https://github.com/gin-gonic/gin/blob/master/go.mod
-go 1.12
+go 1.18
```
iamemilio, line 5: Does this need to be 1.18?

mirackara: I'm not sure if it's _required_, but I switched it in order to match the go version in the [gin/go.mod](https://github.com/gin-gonic/gin/blob/master/go.mod) file.

iamemilio: Is that version of gin the version that is required for it to work again? If so then this is fine.

mirackara: Did some testing and it looks like we actually don't. I will lower the version back to 1.12 now.

nr-swilloughby (approved):

iamemilio (approved):

PR #540, opened 2022-07-25
Harvest Limit Increase + Patch for DT Spans

mirackara:
## Harvest Limit Increase
- Users can use `newrelic.ConfigCustomInsightsEventsMaxSamplesStored()` to change the Harvest Limit to their choosing
- Increased Default Custom Events Harvest Limit from 10k to 30k
- Set a Max Custom Events Harvest Limit at 100k
- Changed static values of samples being stored to dynamic value 
## DT Spans Patch

-  Changed DT Spans being read from `event_harvest_config` to `span_event_harvest_config`

### DIFF
--- a/v3/internal/limits.go
+++ b/v3/internal/limits.go
@@ -13,10 +13,10 @@ const (
 	// MaxPayloadSizeInBytes specifies the maximum payload size in bytes that
 	// should be sent to any endpoint
 	MaxPayloadSizeInBytes = 1000 * 1000
-
 	// MaxCustomEvents is the maximum number of Transaction Events that can be captured
 	// per 60-second harvest cycle
-	MaxCustomEvents = 10 * 1000
+	MaxCustomEvents = 30 * 1000
+	// MaxCustomEvents
 	// MaxLogEvents is the maximum number of Log Events that can be captured per
 	// 60-second harvest cycle
 	MaxLogEvents = 10 * 1000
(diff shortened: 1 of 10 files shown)


iamemilio (approved):

v3/internal/limits.go
```diff
@@ -13,10 +13,9 @@ const (
 	// MaxPayloadSizeInBytes specifies the maximum payload size in bytes that
 	// should be sent to any endpoint
 	MaxPayloadSizeInBytes = 1000 * 1000
-
 	// MaxCustomEvents is the maximum number of Transaction Events that can be captured
 	// per 60-second harvest cycle
-	MaxCustomEvents = 10 * 1000
+	MaxCustomEvents = 30 * 1000
```
iamemilio, line 18: Are we ready to increase the max number of custom events to 30,000 right now?

iamemilio (commented):
looks good, just a question about where we stand with harvest limits currently

v3/internal/limits.go
mirackara: We could hold it until performance testing is done, or we can keep it in. Regardless, the max custom events are still going to be server limited to 10,000.

iamemilio: I think that we should keep defaults the same as they historically have been, just in case things change on the server side. We want to be very explicitly in control of when we expose customers to those kinds of changes.

mirackara: Got it. I will revert that change back.

iamemilio (approved):

PR #548, opened 2022-08-05
Log Writer: an io.Writer that automatically enriches your logs

iamemilio:
This is a go io.Writer that can be used with a number of tested/supported logging frameworks. It can easily be extended to support other logging frameworks as long as they write their outputs to an io.Writer. This tool will append New Relic linking metadata to the end of any log written with it.


iamemilio:
# New Relic Log Writer

 The nrwriter library is an `io.Writer` that can be passed to a number of supported loging frameworks to use New Relic logs in context log enrichment. This writer takes in a log line written by a logging framework, and appends linking metadata to it that associates it to a an application, and when relevant a transaction and span.

 ## Usage

 Once your New Relic application has been created, create an nrwriter instance. It must be passed an io.Writer, which is where the final log content will be written to, and a pointer to New Relic application.

 ```go
 logWriter := nrwriter.New(os.Stdout, app)
 ```

 If any errors occor while trying to decorate your log with New Relic metadata, it will fail silently and print your log message in its original form by default. If you want to see the error messages, then enable debug logging. This will print an error message in a new line after the original log message is printed.

 ```go
 logWriter.DebugLogging(true)
 ```

 Two methods have been created to add transaction data to your nrwriter. To avoid complexity, these functions will make new copy of your nrlogger for that given transaction. You should always make a new copy of your log writer updated with the new transaction nrwriter copy, rather than updating the current log writer's output, otherwise it is possible for concurrent threads to get linked to transactions they do not belong to.

 If you have a pointer to a transaction, use the `WithTransaction()` function. 

 ```go
 txn := app.StartTransaction("my transaction")
 defer txn.End()
 txnWriter := logWriter.WithTransaction(txn)
 ```

 If you have a context with a transaction pointer in it, use the `WithContext()` function. 

 ```go
 func ExampleHandler(w http.ResponseWriter, r *http.Request) {
     txnWriter := logWriter.WithContext(r.Context())
 }
 ```

 ## Supported Frameworks

 The following frameworks are supported:

 - Zerolog
 - Log (go standard library)
 - Logrus

v3/integrations/logcontext-v2/nrwriter/example/main.go
```diff
@@ -0,0 +1,48 @@
...
+		newrelic.ConfigInfoLogger(os.Stdout),
+		newrelic.ConfigAppLogDecoratingEnabled(true),
+	)
+	if err != nil {
+		panic(err)
+	}
+
+	app.WaitForConnection(5 * time.Second)
+
+	nrwriter := nrwriter.New(os.Stdout, app)
```
nr-swilloughby, line 29: I can see it might be a little confusing to make a local variable with the same name as the `nrwriter` package.

nr-swilloughby (dismissed):

iamemilio:
Let's hold this until after recharge. I would like to think about how this could incorporate metrics and maybe auto ingestion too.

iamemilio:
This has been updated as a one stop shop solution. All you need is the zerologWriter package to capture log metrics, events, and to decorate local logs.

iamemilio:
Requires https://github.com/newrelic/go-agent/pull/562 to be merged after to enable testing. May require additional modification of go mod files for frameworks. Do not merge this to master until https://github.com/newrelic/go-agent/pull/562 is also merged.

nr-swilloughby (dismissed):

nr-swilloughby (approved):

v3/integrations/logcontext-v2/zerologWriter/zerolog-writer.go
```diff
@@ -0,0 +1,123 @@
+package zerologWriter
+
+import (
+	"context"
+	"io"
+	"strings"
+
+	"github.com/newrelic/go-agent/v3/internal/logcontext/nrwriter"
```
Meroje, line 8: the fact that logcontext-v2 was made it's own module means the internal from newrelic/v3 can't be imported ..

iamemilio: @Meroje what do you mean?

Meroje: Sorry, I had gone from the initial error I got from failing imports.

There is a missing import for `"github.com/newrelic/go-agent/v3/internal"`, and the capitalization is mismatched for `"github.com/newrelic/go-agent/v3/internal/logcontext/nrwriter"` which prevents it from being found.

iamemilio: @Meroje the problem is actually that we created a submodule in internal :(. Looks like we will have to hotfix this.

iamemilio: I'm pretty sure that most of the stuff in internal is not package correctly to be protected from external imports haha. We will have a fix for this immediately :)

PR #553, opened 2022-08-17
Default Custom Event Harvest Increase from 10,000 to 30,000 

mirackara:
To help our customers see more of the custom events they're producing, the default Custom Events reservoir limit has been increased from 10,000 events per minute to 30,000 events per minute.


RichVanderwal:
We're thinking this needs to be customer configurable, adding a setting here: https://github.com/newrelic/go-agent/blob/06c801d5571056abac8ac9dfa07cf12ca869e920/v3/newrelic/config.go#L69-L74

nr-swilloughby (approved):

nr-swilloughby (approved):

Issue #557, opened 2022-08-19
massive CPU usage at scale with v3.18.0


rittneje:
After upgrading our services to v3.18.0 from v3.17.0, we noticed a massive increase in CPU usage (to almost 100%) in our services that have the most production load at scale. We do not have code-level metrics enabled. We did not change our configuration at all from v3.17.0.

rittneje:
I grabbed a cpu profile while attempting to reproduce the issue. It suggests the issue lies here:
https://github.com/newrelic/go-agent/blob/5421539b6c7c5d481958236cf247759eff0700ad/v3/newrelic/instrumentation.go#L47

Basically, code-level metrics do not perform well at scale due to the caller lookups. And `WrapHandle` is always fetching this information, even if CLM is disabled for the entire app. I don't know why `WithThisCodeLocation()` is being forcefully added instead of making the caller pass it in, but at the very least it should not be injected if the app has CLM disabled.

nr-swilloughby:
The `WithThisCodeLocation()` call is there to accurately identify the wrapping function as the function instead of looking elsewhere, although the user may override that if they choose. However, you are right, that shouldn't be called if CLM is disabled.

rittneje:
@nr-swilloughby Even if we did have CLM enabled, that doesn't mean we want it for every single handler. But if `WrapHandle` always calls `WithThisCodeLocation` then there is no way to conditionally skip it.

nr-swilloughby:
Yeah I see that. I'm thinking I can make a simple change that would make the call to `WithThisCodeLocation` conditional on at least CLM being disabled, as well as allowing you to select CLM on/off for specific wrapped functions. Would it be useful to make wrapped functions be an entire class of things that can be selected on/off for CLM? So you could enable CLM except for wrapped functions if you wanted?

nr-swilloughby:
...and if you don't collect CLM for a particular wrapped function I'll suppress the call to `WithThisCodeLocation` as well.

rittneje:
> Would it be useful to make wrapped functions be an entire class of things that can be selected on/off for CLM?

I didn't quite understand this. Do you mean having another application flag for whether `WrapHandle` and `WrapHandler` should auto-inject `WithThisCodeLocation`? If so I think that's reasonable.

> if you don't collect CLM for a particular wrapped function I'll suppress the call to WithThisCodeLocation as well.

[...]

PR #558, opened 2022-08-24
Corrections to CLM data collection code

nr-swilloughby:
Updates and corrects CLM collection code.

Fixes [issue 557](https://github.com/newrelic/go-agent/issues/557) by correcting where we were unconditionally calling functions to resolve the code location even when code level metrics were disabled globally, as well as doing redundant work in some cases if the user passed their own code locations in. There was also a race condition and thread incompatibility for wrapped function handlers that has been corrected.

Now the agent will never try to resolve code locations unless explicitly asked to (e.g., by the user directly calling `ThisCodeLocation` or similar function), until it knows for sure it needs to collect CLM for that transaction. It will also look to see if code level metrics are locally suppressed (`WithoutCodeLevelMetrics()`) or disabled globally and avoid going through the overhead of collecting anything in those cases. It will also avoid adding its own calls to `ThisCodeLocation` or `WithFunctionLocation` (etc) if the user supplied their own explicit function location in the option list. (Previously, it would call one or both of those and discard the results if overridden downstream by user-set options.)

Additionally, this PR includes some trivial enhancements that were prompted by the above corrections, and/or we expect will be needed once more users start using the CLM feature so it's better to head that off and get them implemented now:

Like we did with `IgnoredPrefix`, we expanded `PathPrefix` from a single `string` value to a new `[]string` field `PathPrefixes` (which deprecates the old `PathPrefix` field). Congruent with that, `ConfigCodeLevelMetricsPathPrefix` now takes any number of `string` values and `NEW_RELIC_CODE_LEVEL_METRICS_PATH_PREFIX` environment variable interprets its value as a comma-separated list of path prefix strings.

A transaction-specific `WithPathPrefix` option allows the specification of a different set of source path prefixes to be given for individual transactions instead of the globally configured list.

A `WithCodeLevelMetrics` option has been added for transactions as the complement to `WithoutCodeLevelMetrics`. This allows for individual transactions to report CLM even if they wouldn't have otherwise (due, for example, to a narrower scope being globally configured which would have excluded this trace). Note that this will *never* enable CLM for any transaction which also carries a `WithoutCodeLevelMetrics` option, or if CLM are disabled globally for the application.

A `WithDefaultFunctionLocation` option has been added which is identical to `WithFunctionLocation` except that it will only _evaluate_ the location of the given function value if no other option has set the code location to be reported before `WithDefaultFunctionLocation` was encountered in the option list.

### DIFF
--- a/v3/newrelic/code_level_metrics.go
+++ b/v3/newrelic/code_level_metrics.go
@@ -73,6 +73,11 @@ func (c *CachedCodeLocation) IsValid() bool {
 	return c != nil && c.once != nil
 }
 
+//
+// NewCachedCodeLocation returns a pointer to a newly-created
+// CachedCodeLocation value, suitable for use with the methods
+// defined for that type.
+//
 func NewCachedCodeLocation() *CachedCodeLocation {
 	return &CachedCodeLocation{
 		once: new(sync.Once),
--- a/v3/newrelic/code_level_metrics.go
+++ b/v3/newrelic/code_level_metrics.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"runtime"
 	"strings"
+	"sync"
 )
 
 //
@@ -33,6 +34,22 @@ type CodeLocation struct {
 	FilePath string
 }
 
+//
+// CachedCodeLocation provides storage for the code location computed such that
+// the discovery of the code location is only done once; thereafter the cached
+// value is available for use.
+//
+// This type includes methods with the same names as some of the basic code location
+// reporting functions and TraceOptions. However, when called as methods of a CachedCodeLocation value
+// instead of a stand-alone function, the operation will make use of the cache to
+// prevent computing the same source location more than once.
+//
+type CachedCodeLocation struct {
+	Location *CodeLocation
+	Err      error
+	once     sync.Once
+}
+
 type traceOptSet struct {
 	LocationOverride *CodeLocation
 	SuppressCLM      bool
@@ -43,6 +60,9 @@ type traceOptSet struct {
 
 //
 // TraceOption values provide optional parameters to transactions.
+//
+// (Currently it's only implemented for transactions, but the name TraceOption is
+// intentionally generic in case we apply these to other kinds of traces in the future.)
 //
 type TraceOption func(*traceOptSet)
 
@@ -108,8 +128,8 @@ func WithoutCodeLevelMetrics() TraceOption {
 // WithCodeLevelMetrics includes this trace in code level metrics even if
 // it would otherwise not be (for example, if it would be out of the configured
 // scope setting). This will never cause code level metrics to be reported if
-// CLM were explicitly disabled (e.g. by CLM being globally off or WithoutCodeLevelMetrics
-// being present in the options for this trace).
+// CLM were explicitly disabled (e.g. by CLM being globally off or if WithoutCodeLevelMetrics
+// is present in the options for this trace).
 //
 func WithCodeLevelMetrics() TraceOption {
 	return func(o *traceOptSet) {
@@ -127,6 +147,16 @@ func WithCodeLevelMetrics() TraceOption {
 //
 func WithThisCodeLocation() TraceOption {
 	return WithCodeLocation(ThisCodeLocation(1))
+}
+
+//
+// WithThisCodeLocation is equivalent to the standalone WithThisCodeLocation
+// TraceOption, but uses the cached value in its receiver to ensure that the
+// overhead of computing the code location is only performed the first time
+// it is invoked for each instance of the receiver variable.
+//
+func (c *CachedCodeLocation) WithThisCodeLocation() TraceOption {
+	return WithCodeLocation(c.ThisCodeLocation(1))
 }
 
 //
@@ -157,6 +187,27 @@ func FunctionLocation(function interface{}) (*CodeLocation, error) {
 }
 
 //
+// FunctionLocation works identically to the stand-alone FunctionLocation function,
+// in that it determines the souce code location of the named function, returning
+// a pointer to a CodeLocation value which represents that location, or an error value
+// if it was unable to find a valid code location for the provided value. However,
+// unlike the stand-alone function, this stores the result in the CachedCodeLocation receiver;
+// thus, subsequent invocations of FunctionLocation for the same receiver will result in
+// immediately repeating the value (or error, if applicable) obtained from the first
+// invocation.
+//
+// This is thread-safe and is intended to allow the same code to run in multiple
+// concurrent goroutines without needlessly recalculating the location of the
+// function value.
+//
+func (c *CachedCodeLocation) FunctionLocation(function interface{}) (*CodeLocation, error) {
+	c.once.Do(func() {
+		c.Location, c.Err = FunctionLocation(function)
+	})
+	return c.Location, c.Err
+}
+
+//
 // WithFunctionLocation is like WithThisCodeLocation, but uses the
 // function value passed as the location to report. Unlike FunctionLocation,
 // this does not report errors explicitly. If it is unable to use the
@@ -165,6 +216,25 @@ func FunctionLocation(function interface{}) (*CodeLocation, error) {
 func WithFunctionLocation(function interface{}) TraceOption {
 	return func(o *traceOptSet) {
 		loc, err := FunctionLocation(function)
+		if err == nil {
+			o.LocationOverride = loc
+		}
+	}
+}
+
+//
+// WithFunctionLocation works like the standalone function WithFunctionLocation,
+// but it stores a copy of the function's location in its receiver the first time
+// it is used. Subsequently that cached value will be used instead of computing
+// the source code location every time.
+//
+// This is thread-safe and is intended to allow the same code to run in multiple
+// concurrent goroutines without needlessly recalculating the location of the
+// function value.
+//
+func (c *CachedCodeLocation) WithFunctionLocation(function interface{}) TraceOption {
+	return func(o *traceOptSet) {
+		loc, err := c.FunctionLocation(function)
 		if err == nil {
 			o.LocationOverride = loc
 		}
@@ -184,6 +254,39 @@ func WithDefaultFunctionLocation(function interface{}) TraceOption {
 	return func(o *traceOptSet) {
 		if o.LocationOverride == nil {
 			WithFunctionLocation(function)(o)
+		}
+	}
+}
+
+//
+// WithDefaultFunctionLocation works like the standalone WithDefaultFunctionLocation function,
+// except that it takes a CachedCodeLocation receiver which will
+// be used to cache the source code location of the function value.
+//
+// Thus, this will arrange for the given function to be reported in Code Level Metrics
+// only if no other option that came before it gave an explicit location to use instead,
+// but will also cache that answer in the provided CachedCodeLocation receiver variable, so that
+// if called again with the same CachedCodeLocation variable, it will avoid the overhead
+// of finding the function's location again, using instead the cached answer.
+//
+// This is thread-safe and is intended to allow the same code to run in multiple
+// concurrent goroutines without needlessly recalculating the location of the
+// function value.
+//
+// If an error is encountered when trying to evaluate the source code location of
+// the provided function value, WithCachedDefaultFunctionLocation will not set anything
+// for the reported code location, and the error will be available as a non-nil value
+// in the Err member of the CachedCodeLocation variable.
+// In this case, no additional attempts are guaranteed to be made on subsequent executions
+// to determine the code location.
+//
+func (c *CachedCodeLocation) WithDefaultFunctionLocation(function interface{}) TraceOption {
+	return func(o *traceOptSet) {
+		if o.LocationOverride == nil {
+			loc, err := c.FunctionLocation(function)
+			if err == nil {
+				WithCodeLocation(loc)(o)
+			}
 		}
 	}
 }
@@ -238,6 +341,34 @@ func ThisCodeLocation(skipLevels ...int) *CodeLocation {
 		loc.FilePath = frame.File
 	}
 	return &loc
+}
+
+//
+// ThisCodeLocation works identically to the stand-alone ThisCodeLocation function,
+// in that it determines the souce code location from whence it was called, returning
+// a pointer to a CodeLocation value which represents that location. However,
+// unlike the stand-alone function, this stores the result in the CachedCodeLocation receiver;
+// thus, subsequent invocations of ThisCodeLocation for the same receiver will result in
+// immediately repeating the value obtained from the first
+// invocation.
+//
+// This is thread-safe and is intended to allow the same code to run in multiple
+// concurrent goroutines without needlessly recalculating the location of the
+// caller.
+//
+func (c *CachedCodeLocation) ThisCodeLocation(skiplevels ...int) *CodeLocation {
+	var skip int
+
+	if len(skiplevels) > 0 {
+		skip = skiplevels[0]
+	}
+
+	c.once.Do(func() {
+		// add 4 skip levels to compensate for the internal calls used to get here.
+		c.Location = ThisCodeLocation(skip + 4)
+		c.Err = nil
+	})
+	return c.Location
 }
 
 func removeCodeLevelMetrics(remAttr func(string)) {
--- a/v3/newrelic/code_level_metrics.go
+++ b/v3/newrelic/code_level_metrics.go
@@ -36,7 +36,9 @@ type CodeLocation struct {
 type traceOptSet struct {
 	LocationOverride *CodeLocation
 	SuppressCLM      bool
+	DemandCLM        bool
 	IgnoredPrefixes  []string
+	PathPrefixes     []string
 }
 
 //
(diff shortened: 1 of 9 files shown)


nr-swilloughby:
This is still very much a work in progress but I'm drafting the PR now to make collaboration and review easier along the way, since this is important that we get this released ASAP.

It is still experimental, untested, and subject to changes. I'll be added testcases and doing more full testing next.

One thing that I considered adding but have _not_ yet is the ability to set the global scope to exclude wrapped functions from CLM as a class unto themselves. This would mean you could collect CLM for anything you called `StartTransaction` for (or some integration code did on your behalf) but wrapped handler functions would not automatically get CLM as they do now.

I'm thinking that with the improvements here that wouldn't be a necessary step, but if it does prove to be a useful configuration option I can add that as well by expanding the list of scopes from the current set of
 * `AllCLM`
 * `TransactionCLM`
to
 * `AllCLM`
 * `TransactionCLM` (i.e. the combination of ExplicitTransactionCLM and WrappedTransactionCLM)
 * `ExplicitTransactionCLM` (i.e., normal `StartTransaction` calls)
 * `WrappedTransactionCLM` (functions in `WrapHandle` and `WrapHandleFunc`)
That would necessitate the changing of the raw numeric values behind those constants in order to preserve `TransactionCLM`'s semantic meaning for backward compatibility. I have added a note in the comments for those constants warning against relying on those underlying values anyway, and advising to just use the defined names like `TransactionCLM` instead of using `1` for that value.

v3/newrelic/instrumentation.go
```diff
@@ -36,16 +36,32 @@ func WrapHandle(app *Application, pattern string, handler http.Handler, options
...
+				if tOptions.LocationOverride == nil {
+					WithThisCodeLocation()(tOptions)
+				}
+			}
+		}
+		if tOptions == nil {
+			// we weren't able to curate the options above, so pass whatever we were given downstream
+			txnOptionList = options
+		} else {
+			txnOptionList = append(txnOptionList, withPreparedOptions(tOptions))
```
rittneje, line 61: I think this can be simplified:
```go
txOptionList = []TraceOption{ func(o *traceOptSet) { *o = *tOptions } }
```

v3/newrelic/instrumentation.go
```diff
@@ -36,16 +36,32 @@ func WrapHandle(app *Application, pattern string, handler http.Handler, options
...
+			if tOptions != nil && !tOptions.SuppressCLM && (tOptions.DemandCLM || app.app.run.Config.CodeLevelMetrics.Scope == 0 || (app.app.run.Config.CodeLevelMetrics.Scope&TransactionCLM) != 0) {
+				// we are for sure collecting CLM here, so go to the trouble of collecting this code location.
+				if tOptions.LocationOverride == nil {
+					WithThisCodeLocation()(tOptions)
+				}
+			}
+		}
+		if tOptions == nil {
+			// we weren't able to curate the options above, so pass whatever we were given downstream
+			txnOptionList = options
```
rittneje, line 59: The only way to reach this case is if CLM is disabled for the whole app, in which case I think it doesn't matter whether forward the options or not. In other words I think you can just assign to `txnOptionList` up in the `if tOptions != nil && !tOptions.SuppressCLM && ...` block above and it will work out.

v3/newrelic/code_level_metrics.go
```diff
@@ -142,6 +171,46 @@ func WithFunctionLocation(function interface{}) TraceOption {
 	}
 }
 
+//
+// WithDefaultFunctionLocation is like WithFunctionLocation but will only
+// evaluate the location of the function if nothing that came before it
+// set a code location first. This is useful, for example, if you want to
+// provide a default code location value to be used but not pay the overhead
+// of resolving that location until it's clear that you will need to. This
+// should appear at the end of a TraceOption list (or at least before any
```
rittneje, line 180: I think you meant _after_ any other options that want to specify the code location

v3/newrelic/instrumentation.go
nr-swilloughby: True, but I'm thinking ahead to the case where there may be non-CLM TraceOptions that may be defined at some point. In that case they would still work even if we weren't doing code level metrics. We're still avoiding the unnecessary call to get the function location.

However for the time being it may be fine to leave the code simpler and expand it as appropriate at some future point where there's anything in there other than CLM options.

v3/newrelic/code_level_metrics.go
nr-swilloughby: Er, yes. _after_.

nr-swilloughby:
Moved from the automatically-set "awaiting user input" state to the actually-accurate "in progress" one.

v3/newrelic/instrumentation.go
nr-swilloughby: At the moment that's equivalent and a tiny bit faster to execute, and if I don't end up needing the more nuanced implementation of `withPreparedOptions` which won't do things like overwrite a non-empty slice of values with an empty one from the incoming struct, I'll probably refactor that back to what's suggested here.
And it may likely turn out that the `withPreparedOptions` thing isn't needed. I'm experimenting with a few ideas here and refactoring out what doesn't seem to add value.

v3/newrelic/config.go
```diff
@@ -397,18 +411,22 @@ type Config struct {
...
 //
-// A scope of 0 means "all types" as a convenience so that any zero-value
-// CodeLevelMetricsScope variable provides the default expected behavior
-// rather than turning off all code level metrics.
+// The zero value of CodeLevelMetricsScope means "all types" as a convenience so that
+// new variables of this type provide the default expected behavior
+// rather than, say, turning off all code level metrics as a 0 bit value would otherwise imply.
+// Otherwise the numeric values of these constants are not to be relied
+// upon and are subject to change. Only use the named constant identifiers in
+// your code. We do not recommend saving the raw numeric value of these constants
```
rittneje, line 425: One potential use case to consider is someone trying to configure it through, say, a config file. I see that `NEW_RELIC_CODE_LEVEL_METRICS_SCOPE` is a comma-separated list of strings, but the corresponding `codeLevelMetricsScopeLabelToValue` function is private.

nr-swilloughby: It might be better to provide a fully-functional function like `ConfigFromEnvironment()` that would read all the configuration values from a file or a `map[string]string` vlaue that had been read from a config file and set them up in the agent like `ConfigFromEnvironment()` does, saving the work of custom code to take the values read from the configuration file and do the right thing for all of them.

v3/newrelic/instrumentation.go
```diff
@@ -36,16 +36,32 @@ func WrapHandle(app *Application, pattern string, handler http.Handler, options
...
-		txn := app.StartTransaction(r.Method+" "+pattern, options...)
+		var tOptions *traceOptSet
+		var txnOptionList []TraceOption
+
+		if app.app != nil && app.app.run != nil && app.app.run.Config.CodeLevelMetrics.Enabled {
+			tOptions = resolveCLMTraceOptions(options)
+			if tOptions != nil && !tOptions.SuppressCLM && (tOptions.DemandCLM || app.app.run.Config.CodeLevelMetrics.Scope == 0 || (app.app.run.Config.CodeLevelMetrics.Scope&TransactionCLM) != 0) {
+				// we are for sure collecting CLM here, so go to the trouble of collecting this code location.
+				if tOptions.LocationOverride == nil {
+					WithThisCodeLocation()(tOptions)
```
rittneje, line 53: I think the bigger issue was not the lookup itself, but rather the fact that it was happening from scratch on each transaction, even though the answer should always be the same. So maybe it is sufficient to move the lookup outside the closure and have it based on `handler.ServeHTTP`.
Actually that would be a departure from the current behavior, which seems to be looking for the caller of the _wrapped_ handler. I'm not sure which of the two is truly intended. But as things currently stand it seems to be a discrepancy between `WrapHandle` and `WrapHandler`.

nr-swilloughby: The intent is to report the code location where the transaction occurs. In the case where we are passed a `func` value that's easy, it's the first line of that function.

nr-swilloughby: Avoiding repeatedly calling it every time the closure is invoked would be preferable, though, as long as the code location is accurately indicating where the transaction is. Perhaps a one-time lookup with a cached answer...?

rittneje: I think if you literally just do `options = append([]TraceOption{WithFunctionLocation(handler.ServeHTTP)}, options...)` _outside_ the closure it will resolve all the issues.

nr-swilloughby: I have an idea to address that too. I'll work up a proof of concept.

rittneje: @nr-swilloughby Please see #561 for my proposed fixes.

nr-swilloughby: The solution I have in progress goes a bit further to ensure that we avoid ever looking up source code locations we don't need, even once, in addition to resolving the concurrency issue, and includes providing that functionality for user-supplied code that needs to find source locations in performance-critical and/or concurrent code that isn't just in the wrapped functions.

However, I think having WrapHandle report `handler.ServeHTTP` instead of the internal closure as the code location to report makes more sense, so I'm making that change as well.

v3/newrelic/config.go
nr-swilloughby: I created a feature request separate from this to consider providing support for config files and whatnot, but I think it would make sense to implement `encoding.TextUnmarshaler` for `CodeLevelMetricsScope` makes sense for this PR on its own.

v3/newrelic/code_level_metrics.go
```diff
@@ -33,15 +34,36 @@ type CodeLocation struct {
...
+//
+// This type includes methods with the same names as some of the basic code location
+// reporting functions and TraceOptions. However, when called as methods of a CachedCodeLocation value
+// instead of a stand-alone function, the operation will make use of the cache to
+// prevent computing the same source location more than once.
+//
+type CachedCodeLocation struct {
+	Location *CodeLocation
+	Err      error
+	once     sync.Once
```
rittneje, line 50: A `sync.Once` cannot be copied after first use. Unless you use a pointer here, then if someone copies a `CachedCodeLocation` around it will cause issues.

Also, I think `Location` and `Err` need to be private. There is no way for package clients to use them safely, nor is there a reason for them to do so.

v3/newrelic/code_level_metrics.go
```diff
@@ -171,13 +343,58 @@ func ThisCodeLocation(skipLevels ...int) *CodeLocation {
...
+//
+func (c *CachedCodeLocation) ThisCodeLocation(skiplevels ...int) *CodeLocation {
+	var skip int
+
+	if len(skiplevels) > 0 {
+		skip = skiplevels[0]
+	}
+
+	c.once.Do(func() {
+		// add 4 skip levels to compensate for the internal calls used to get here.
```
rittneje, line 367: This is contingent upon the implementation of the sync package, which could be a problem.

v3/newrelic/config.go
```diff
@@ -397,29 +411,102 @@ type Config struct {
...
 	}
-	return 0, false
+	return scope, ok
+}
+
+//
+// UnmarshalJSON allows for a CodeLevelMetricsScope value to be read from a JSON
+// string whose value is a comma-separated list of scope labels.
+//
+func (s *CodeLevelMetricsScope) UnmarshalJSON(b []byte) error {
```
rittneje, line 472: Please implement `UmarshalText` instead as it is more generic. (It will be used by JSON, XML, and others.)

v3/newrelic/code_level_metrics.go
nr-swilloughby: Making the `once` value a pointer shouldn't be a problem. I also agree with the suggestion to make the members private here. Really the whole `CachecCodeLocation` value should be opaque to the caller.

v3/newrelic/code_level_metrics.go
nr-swilloughby: What issues do you see with the sync package? I thought that was a fairly well-defined part of the standard library at this point. But if there are issues we could explicitly document expectations around that or do something to mitigate the issues.

v3/newrelic/instrumentation.go
```diff
@@ -36,16 +36,40 @@ func WrapHandle(app *Application, pattern string, handler http.Handler, options
...
-		txn := app.StartTransaction(r.Method+" "+pattern, options...)
+		var tOptions *traceOptSet
+		var txnOptionList []TraceOption
+
+		if app.app != nil && app.app.run != nil && app.app.run.Config.CodeLevelMetrics.Enabled {
+			tOptions = resolveCLMTraceOptions(options)
+			if tOptions != nil && !tOptions.SuppressCLM && (tOptions.DemandCLM || app.app.run.Config.CodeLevelMetrics.Scope == 0 || (app.app.run.Config.CodeLevelMetrics.Scope&TransactionCLM) != 0) {
+				// we are for sure collecting CLM here, so go to the trouble of collecting this code location if nothing else has yet.
+				if tOptions.LocationOverride == nil {
+					if loc, err := cache.FunctionLocation(handler); err == nil {
```
rittneje, line 55: You can simplify away the need for two caches if you instead put the fallback logic into the callback that the single cache will invoke within the `sync.Once`.

v3/newrelic/code_level_metrics.go
rittneje: The fact that there are 4 levels is an implementation detail. If the Go authors refactor the way `sync.Once.Do` works internally it could change.

nr-swilloughby: Oh, that. Yes, that's on my mind as well. I'm afraid there may need to be some more logic in there to do something more intelligent than blindly skipping a number of levels, like skipping over things that belong to the sync package.

v3/newrelic/config.go
nr-swilloughby: That was trivial enough that I just pushed the change for it. I'll look at the other suggestions next.

v3/newrelic/code_level_metrics.go
nr-swilloughby: Ok, I'm experimenting now with a new `CachedCodeLocation` type which uses a pointer to the `once` field so copying the struct should just give you another struct that tracks to the same `once` value instead of copying it, and has the fields all unexported. That does mean you need to call `NewCachedCodeLocation()` to get one with the pointer set up for you instead of just making a zero-value `CachedCodeLocation` so that the pointer is set up before you start doing anything concurrent in there.

v3/newrelic/instrumentation.go
nr-swilloughby: I refactored that to only need one cache value.

v3/newrelic/code_level_metrics.go
nr-swilloughby: The updated code just pushed implements this approach.

v3/newrelic/code_level_metrics.go
nr-swilloughby: The code just pushed includes a new implementation that more intentionally steps around the internal agent code and calls out to the sync library and back rather than just skipping a number of calls, which should mitigate this issue.

v3/newrelic/code_level_metrics_test.go
```diff
@@ -15,15 +16,15 @@ func anotherFunction() {
 
 func TestCodeLocation(t *testing.T) {
 	loc1 := ThisCodeLocation()
-	if loc1.LineNo != 17 || loc1.Function != "github.com/newrelic/go-agent/v3/newrelic.TestCodeLocation" || !strings.HasSuffix(loc1.FilePath, "/go-agent/v3/newrelic/code_level_metrics_test.go") {
+	if loc1.LineNo != 18 || loc1.Function != "github.com/newrelic/go-agent/v3/newrelic.TestCodeLocation" || !strings.HasSuffix(loc1.FilePath, "/go-agent/v3/newrelic/code_level_metrics_test.go") {
```
iamemilio, line 19: what does this mean? Why can it not be line number 18?

`loc1.LineNo != 18`

v3/newrelic/code_level_metrics.go
```diff
@@ -33,15 +34,70 @@ type CodeLocation struct {
...
+// reporting functions and TraceOptions. However, when called as methods of a CachedCodeLocation value
+// instead of a stand-alone function, the operation will make use of the cache to
+// prevent computing the same source location more than once.
+//
+// A usable CachedCodeLocation value must be obtained via a call to
+// NewCachedCodeLocation.
+//
+type CachedCodeLocation struct {
+	location *CodeLocation
+	once     *sync.Once
```
rittneje, line 52: Consider using `atomic.LoadPointer`, etc. instead of `sync.Once`. While it won't be air-tight (two goroutines might do the same lookup in parallel) it should be good enough for what you need, and avoids the extra logic you had to add to omit sync from the stack frame lookup.

iamemilio (approved):
Looks good to me. We will do a little more end to end testing just to verify before merging to master.

v3/newrelic/code_level_metrics.go
nr-swilloughby: Good suggestion, and I considered that earlier on, but I think it would be better to go to the effort to use the more robust `sync.Once` function to avoid possible parallel lookups and such, and I think the way the code now steps around the calls to whatever's in the middle (e.g. `sync.Once`) will be flexible enough to support future needs as well.

iamemilio: I think we would rather be defensive here, but appreciate the suggestion

PR #567, opened 2022-09-12
Unifi sysinfo/memtotal for all openbsd arches.

qbit:
On OpenBSD syscalls are the same across arches so there is no need to limit it to only amd64.


Ak-x:
Hello @qbit, we require a CLA signature before we can proceed (and for this code to be rebased).  Once that is complete we can proceed.

ninckblokje:
Hi all,

Sorry to post on this old PR, but I was wondering if it would be possible to re-open this PR and see if the PR can be merged?

jcourreges:
> Hello @qbit, we require a CLA signature before we can proceed (and for this code to be rebased). Once that is complete we can proceed.

@Ak-x could you please reopen this?  As @qbit [mentioned] he had already signed the CLA.

As this module starts to get used in other projects, the limitation to amd64 becomes more of a problem.

See https://github.com/openbsd/ports/commit/97b058d977471c59985d1f419b94f4726b84c10a

PR #577, opened 2022-09-17
updated nrgin integration to report code level metrics correctly

nr-swilloughby:
Implements correct handler function location when using the nrgin integration with Code-Level Metrics enabled.


iamemilio (approved):
nice and simple. Looks good to me

PR #590, opened 2022-10-19
Add Module Dependency Metrics to the Go Agent

nr-swilloughby:
This implements module dependency metrics in the Go Agent. This feature reports the application's dependencies when connecting to the New Relic service, allowing the customer to better manage dependencies and be aware of security vulnerabilities in libraries they use.

This version only collects the name and version number for each imported package.

Also introduces the ability to redact what modules were explicitly excluded by the user, and adds that capability to the Code Level Metrics capability as well.

### DIFF
--- a/v3/newrelic/config.go
+++ b/v3/newrelic/config.go
@@ -642,10 +642,13 @@ func defaultConfig() Config {
 
 	// Code Level Metrics
 	c.CodeLevelMetrics.Enabled = false
+	c.CodeLevelMetrics.RedactPathPrefixes = true
+	c.CodeLevelMetrics.RedactIgnoredPrefixes = true
 	c.CodeLevelMetrics.Scope = AllCLM
 
 	// Module Dependency Metrics
 	c.ModuleDependencyMetrics.Enabled = true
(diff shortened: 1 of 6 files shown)


v3/newrelic/config.go
```diff
@@ -401,21 +407,34 @@ type Config struct {
...
 	}
+
+	// ModuleDependencyMetrics controls reporting of the packages used to build the instrumented
+	// application, to help manage project dependencies.
+	ModuleDependencyMetrics struct {
+		// Enabled controls whether the module dependencies are collected and reported.
+		Enabled bool
+		// RedactIgnoredPrefixes, if true, redacts a non-nil list of IgnoredPrefixes from
+		// the configuration data transmitted by the agent.
+		RedactIgnoredPrefixes bool
```
iamemilio, line 418: IMO this should just be the default behavior, or default to true, especially since there isn't an easy way to set this in the config_options. Alternatively, if the customer is running in high security mode, we could always set this to true.

nr-swilloughby: Oops. I meant to provide options to set those. I also agree that making them on by default makes sense.

iamemilio (approved):

PR #599, opened 2022-11-04
nrawssdk-v2: fix usage examples of nrawssdk.AppendMiddlewares

Meroje:
## Details

When testing the nrawssdk-v2 integration along with nrlambda I found out that the provided method of injecting the middleware to APIOptions does not work (no trace is seen for any aws call), what worked was to use the optFns param.

In this PR I'm only changing the documentation and example to reflect what has been working, but maybe you'd want to expand on it to add an exported function that can be used as an optFn ?

Also I could not find how to make the test fail as I'm not familiar enough with the internals.

### DIFF
--- a/v3/integrations/nrawssdk-v2/example/main.go
+++ b/v3/integrations/nrawssdk-v2/example/main.go
@@ -10,7 +10,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
-	nraws "github.com/newrelic/go-agent/v3/integrations/nrawssdk-v2"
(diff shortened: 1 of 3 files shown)


iamemilio:
@Meroje Hi, I'm taking a look at this and was wondering what this really changes other than when `nrawssdk.AppendMiddlewares(&awsConfig.APIOptions, nil)` is invoked.

Meroje:
I found that the previous method had no effect and I would never get any telemetry from aws sdk.

iamemilio:
Oh ok, that's a good reason haha. Ok, I'll verify this and see if we can land it in an upcoming release.

nr-swilloughby (approved):
These changes are approved and we're reviewing what potential changes to the integration may be needed to better support AWS as highlighted by this PR.

PR #612, opened 2022-12-10
Update and Fixed Labstack Echo Open Redirect 

iot-defcon:
### 👾 Describe The Sumarry:
Affected of this `newrelic/go-agent` are vulnerable to Open Redirect in handlers serving static directories via the `echo.StaticDirectoryHandler` handler. Labstack Echo v4.8.0 was discovered to contain an open redirect vulnerability via the Static Handler component. This vulnerability can be leveraged by attackers to cause a Server-Side Request Forgery (SSRF). 

```js
		p = c.Request().URL.Path // path must not be empty.
		if fi.IsDir() && p[len(p)-1] != '/' {
		if fi.IsDir() && len(p) > 0 && p[len(p)-1] != '/' {
			// Redirect to ends with "/"
			return c.Redirect(http.StatusMovedPermanently, p+"/")
			return c.Redirect(http.StatusMovedPermanently, sanitizeURI(p+"/"))
		}
		return c.File(name)
	}
	// Handle added routes based on trailing slash:
	// 	/prefix  => exact route "/prefix" + any route "/prefix/*"
	// 	/prefix/ => only any route "/prefix/*"
	if prefix != "" {
		if prefix[len(prefix)-1] == '/' {
			// Only add any route for intentional trailing slash
			return get(prefix+"*", h)
```

### Remendation References
[Nuclei Templates](https://github.com/projectdiscovery/nuclei-templates/blob/master/cves/2022/CVE-2022-40083.yaml)

### 🥷 According CVeScores:
CVE-2022-40083
[CWE-601](https://cwe.mitre.org/data/definitions/601.html)
CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:C/C:H/I:H/A:H

### DIFF
--- a/v3/integrations/nrecho-v4/go.mod
+++ b/v3/integrations/nrecho-v4/go.mod
@@ -8,19 +8,3 @@ require (
 	github.com/labstack/echo/v4 v4.9.0
 	github.com/newrelic/go-agent/v3 v3.18.2
 )
-
-require (
-	github.com/golang/protobuf v1.4.3 // indirect
-	github.com/labstack/gommon v0.3.1 // indirect
-	github.com/mattn/go-colorable v0.1.11 // indirect
-	github.com/mattn/go-isatty v0.0.14 // indirect
-	github.com/valyala/bytebufferpool v1.0.0 // indirect
-	github.com/valyala/fasttemplate v1.2.1 // indirect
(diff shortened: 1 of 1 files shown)


iamemilio:
@iot-defcon would you mind removing the indirect dependencies from the `go.mod` file? We prefer to let the customers generate those in their environments.

iamemilio (approved):

Issue #619, opened 2023-01-02
Compression Memory Leak


mhmtszr:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description
We have been observing the memory increase in our Go applications for a long time. When We look at the cumulative increase of the memories used by the application with the memory dump method, We observed that the New Relic agent uses a lot of memory while **doing gzip compression.**

## Versions
github.com/newrelic/go-agent/v3 v3.20.1
github.com/newrelic/go-agent/v3/integrations/nrgin v1.1.3

## Example memory dump
![image](https://user-images.githubusercontent.com/35079195/210282500-cf7fbbf9-3fe9-4047-a9c3-615d08f72c96.png)

## Details
Instead of creating a new gzip writer for every collect request, use common ones by using gzip writer [sync pool](https://pkg.go.dev/sync#Pool).

iamemilio:
This is great to see, thanks for filing this and a pull request with us! There are a lot of old bits and pieces in the code, and it can be hard for us to spot issues like this. I am not sure I would characterize this as a memory leak from the data you posted, it looks like it's fairly fixed. I would love to see what the memory consumption stats look like with the PR you proposed too.

mhmtszr:
@iamemilio I tested with my PR, and I am sharing the results with you. I only changed the New Relic Go Agent, and now the memory increase decreased by 22%, at the same time, the CPU usage has a noticeable decrease.

Ak-x:
This fix will be included in an upcoming release

mhmtszr:
Fixed with 3.20.3.

PR #620, opened 2023-01-03
Fix memory leak while compression

mhmtszr:
## Links
We have been observing the memory increase in our Go applications for a long time. When We look at the cumulative increase of the memories used by the application with the memory dump method, We observed that the New Relic agent uses a lot of memory while doing gzip compression.

As a result, too much GC is running. You can visualize it by closing GOGC=off and defining GOMEMLIMIT this will avoid unnecessary GC, and will trigger the GC only if it's closed to the GOMEMLIMIT.

Issue: https://github.com/newrelic/go-agent/issues/619

## Details
Instead of creating a new gzip writer for every collect request, use common ones by using gzip writer sync pool.

### DIFF
--- a/v3/newrelic/collector.go
+++ b/v3/newrelic/collector.go
@@ -55,6 +55,7 @@ type rpmControls struct {
 	License string
 	Client  *http.Client
 	Logger  logger.Logger
+	Writer  *gzip.Writer
 }
 
 // rpmResponse contains a NR endpoint response.
@@ -136,11 +137,10 @@ func rpmURL(cmd rpmCmd, cs rpmControls) string {
 	return u.String()
 }
 
-func compress(b []byte) (*bytes.Buffer, error) {
+func compress(b []byte, w *gzip.Writer) (*bytes.Buffer, error) {
 	var buf bytes.Buffer
-	w := gzip.NewWriter(&buf)
+	w.Reset(&buf)
 	_, err := w.Write(b)
-	w.Close()
 
 	if nil != err {
 		return nil, err
@@ -150,7 +150,7 @@ func compress(b []byte) (*bytes.Buffer, error) {
 }
 
 func collectorRequestInternal(url string, cmd rpmCmd, cs rpmControls) rpmResponse {
-	compressed, err := compress(cmd.Data)
+	compressed, err := compress(cmd.Data, cs.Writer)
 	if nil != err {
 		return rpmResponse{Err: err}
 	}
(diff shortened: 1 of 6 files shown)


v3/newrelic/collector.go
```diff
@@ -136,11 +137,10 @@ func rpmURL(cmd rpmCmd, cs rpmControls) string {
 	return u.String()
 }
 
-func compress(b []byte) (*bytes.Buffer, error) {
+func compress(b []byte, w *gzip.Writer) (*bytes.Buffer, error) {
 	var buf bytes.Buffer
-	w := gzip.NewWriter(&buf)
+	w.Reset(&buf)
 	_, err := w.Write(b)
-	w.Close()
```
mstrYoda, line 143: You need to close gzip writer to read that data properly. It adds required gzip headers.

v3/newrelic/collector.go
```diff
@@ -136,11 +137,10 @@ func rpmURL(cmd rpmCmd, cs rpmControls) string {
 	return u.String()
 }
 
-func compress(b []byte) (*bytes.Buffer, error) {
+func compress(b []byte, w *gzip.Writer) (*bytes.Buffer, error) {
```
mstrYoda, line 140: If you access to a pointer gzip writer concurrently, it will probably cause race condition.

mstrYoda (dismissed):

mhmtszr:
@iamemilio this memory leak affects our applications, could you please review it?

Also, I realized that **newAnalyticsEvents** function creates a lot of analytic events on the heap too. We may focus there too.

iamemilio:
Would you be able to re-factor this for 1.17, that is the minimum supported version of Go that we must support currently so we can't use `any` yet unfortunately.

mhmtszr:
@iamemilio I changed any to interface{} and I tested it works with 1.17 could you check it?

iamemilio:
It looks like some of the tests are failing: https://github.com/newrelic/go-agent/actions/runs/3886936188/jobs/6669443533

Our CI is a bit chaotic, I have a fix coming through that will clean it up. I would focus on getting this test to pass for now.

mhmtszr:
I thought I solved the failed tests could you run the pipelines again I can't.

iamemilio:
Thanks for fixing this up, and sorry about the confusing CI. This looks good now! We are going to run some scale tests on our end just to verify, but it looks like this will very likely make it into the next release! Thanks for contributing.

iamemilio (approved):
Confirming a noticeable decrease in CPU and Memory usage

PR #633, opened 2023-01-27
Error Event Data Truncation

mirackara:
Limit error event messages to 4096 bytes in order to prevent transmission of data larger than what the collector would accept.

Also verified that other event types are being properly truncated before transmission


iamemilio (approved):
Good job, and thanks for adding the unit test! LGTM

PR #637, opened 2023-02-01
Add SendQueryParameters flag to nrgpx5 to selectively enable sending database query parameters

dacohen:
For applications handling sensitive data, its useful to be able to configure integrations to *not* send database query parameters upstream to New Relic.

## Details

This PR adds a `SendQueryParameters` flag to the `Tracer` struct, which controls whether `QueryParameters` are set on the Segment. By default, this is set to true, to preserve existing functionality.

### DIFF
--- a/v3/integrations/nrpgx5/nrpgx5.go
+++ b/v3/integrations/nrpgx5/nrpgx5.go
@@ -68,8 +68,9 @@ func init() {
 
 type (
 	Tracer struct {
-		BaseSegment newrelic.DatastoreSegment
-		ParseQuery  func(segment *newrelic.DatastoreSegment, query string)
+		BaseSegment         newrelic.DatastoreSegment
+		ParseQuery          func(segment *newrelic.DatastoreSegment, query string)
+		SendQueryParameters bool
 	}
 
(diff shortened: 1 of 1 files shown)


v3/integrations/nrpgx5/nrpgx5.go
```diff
@@ -83,7 +84,8 @@ const (
 
 func NewTracer() *Tracer {
 	return &Tracer{
-		ParseQuery: sqlparse.ParseQuery,
+		ParseQuery:          sqlparse.ParseQuery,
+		SendQueryParameters: true,
```
iamemilio, line 88: could we default this to false? I don't want customers to feel that their sensitive data was exposed unexpectedly, and so I'd like to preserve the default behavior.

dacohen: Just to clarify, the default behavior before this PR was to always send these parameters. This change preserves that for backwards compatibility.

iamemilio: Oh, gotcha. In that case this is fine then.

iamemilio (approved):

Issue #436, opened 2022-01-14
v3/integrations/nrpq - datastore segments should populate ParameterizedQuery field


garciademarina:
## Summary

When using [v3/integrations/nrpq](https://godoc.org/github.com/newrelic/go-agent/v3/integrations/nrpq) the datastore segment is automatically instrumented however `ParameterizedQuery` field is always empty.

## Desired Behaviour

The nrpq integration should also saved the parametrized query (`ParameterizedQuery`) in the datastore segment. 

## Possible Solution

We did a new integration package in order to save `ParameterizedQuery` automatically by using a custom/modified `parse.ParseQuery` function. 
The `parse.ParseQuery` function receives a pointer to the segment and a query as parameters, it populates the table name and the operation (select, insert, ...), because we also have the query passed as a parameter we are saving it there. 

```go
func ParseQuery(segment *newrelic.DatastoreSegment, query string) {
	s := cCommentRegex.ReplaceAllString(query, "")
	s = lineCommentRegex.ReplaceAllString(s, "")
	s = sqlPrefixRegex.ReplaceAllString(s, "")
	op := strings.ToLower(firstWordRegex.FindString(s))
	if rg, ok := sqlOperations[op]; ok {
		segment.Operation = op
		if nil != rg {
			if m := rg.FindStringSubmatch(s); len(m) > 1 {
				segment.Collection = extractTable(m[1])
			}
		}
		segment.ParameterizedQuery = query // <-- save parameterized query 
	}
}
```

## Additional context

We don't get much information about the queries because they are being grouped like `'select' on 'table' using 'Postgres'`.

mirackara:
Hi @garciademarina 

This enhancement request has been around for a while. I'm closing it out for now, but if the issue is still relevant please don't hesitate to reopen a new request!

PR #642, opened 2023-02-10
Add query to ParameterizedQuery

artefactop:
This PR fix the following issue and we are using it for a while with the desired behaviour

## Links

https://github.com/newrelic/go-agent/issues/436

## Details

We don't get much information about the queries because they are being grouped like 'select' on 'table' using 'Postgres'.

### DIFF
--- a/v3/newrelic/sqlparse/sqlparse.go
+++ b/v3/newrelic/sqlparse/sqlparse.go
@@ -60,6 +60,7 @@ func ParseQuery(segment *newrelic.DatastoreSegment, query string) {
 	s = sqlPrefixRegex.ReplaceAllString(s, "")
 	op := strings.ToLower(firstWordRegex.FindString(s))
 	if rg, ok := sqlOperations[op]; ok {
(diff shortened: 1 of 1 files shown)


iamemilio (approved):

iamemilio (changes requested):
Would it be possible to put a gate here that lets you disable this feature by default like this pull request? https://github.com/newrelic/go-agent/pull/637/files

iamemilio:
We will happily merge these changes, but we want a way for them to be configured to be disabled by default first. Some customers will not appreciate data they were not expecting suddenly getting collected.

bhaskar-simpl:
How about documenting that in the changelog?

bhaskar-simpl:
One way I see is to create a separate version for nrpq that supports this. 
Can you quickly align on the solution rather than keeping this blocked forever?

99-deadshot:
@iamemilio please check #677 on priority

iamemilio:
addressed in #745

PR #644, opened 2023-02-15
Fix for DT Spans reservoir limiting

### LINKED ISSUE #643: Distributed Tracer Reservoir Limit Not Being Used
Reopened #600 as instructed since the latest Go Agent (`v3.20.3`) still has this issue:

Instead of the configurable distributed tracer reservoir limit, the `defaultMaxSpanEvents` is being used to determine whether or not to save a span event.

## Description

https://github.com/newrelic/go-agent/blob/dd6e323446a7aff70e700be2bfae97bf899805b0/v3/newrelic/tracing.go#L416

When adding spans to a transaction, after you have added 2000 spans you are no longer able to add anymore, despite setting a higher reservoir limit using `newrelic.ConfigDistributedTracerReservoirLimit()`. This is because the `defaultMaxSpanEvents` constant is being used in the check, which is set at 2000. There is a note in the code expressing the need to replace this functionality.

https://github.com/newrelic/go-agent/blob/dd6e323446a7aff70e700be2bfae97bf899805b0/v3/newrelic/limits.go#L41

## Steps to Reproduce

Try to add more than 2000 spans to a transaction.

## Expected Behavior

I would expect to see all the spans I instrumented in the code, up to the number I passed in to `newrelic.ConfigDistributedTracerReservoirLimit()`.

mirackara:
A fix for the reservoir size containing DT Spans not correctly being changed.

Fixes #643


v3/newrelic/config.go
```diff
@@ -725,6 +725,16 @@ func (c Config) maxTxnEvents() int {
 	return configured
 }
 
+// maxSpanEvents returns the configured maximum number of Span Events if it has been configured
+// and is less than the default maximum; otherwise it returns the default max.
+func (c Config) maxSpanEvents() int {
+	configured := c.DistributedTracer.ReservoirLimit
+	if configured < 0 || configured > internal.MaxSpanEvents {
```
iamemilio, line 732: you may want to confirm this in agent-dev, but I think the expectation is that agents no longer limit data capacity requests. The connect server will do it in the connect response.

mirackara: Good idea, I'll verify this and update the PR if necessary

v3/newrelic/tracing.go
```diff
@@ -413,9 +413,7 @@ func (t *txnData) CurrentSpanIdentifier(thread *tracingThread) string {
 
 func (t *txnData) saveSpanEvent(e *spanEvent) {
 	e.AgentAttributes = t.Attrs.filterSpanAttributes(e.AgentAttributes, destSpan)
-	if len(t.SpanEvents) < defaultMaxSpanEvents {
```
iamemilio, line 416: I am a little surprised there was a hard cap on the span events here, and not an attempt to sample them. That being said, I don't think we should just outright remove this unless we have a very good reason to. It is checking the wrong value though. It should check against the value of run.MaxSpanEvents().

mirackara: As we cannot access the current run in the tracing go file, I set the check against internal.MaxSpanEvents which is the replacement for the deprecated `defaultMaxSpanEvents`

iamemilio (approved):

nr-swilloughby (approved):

PR #706, opened 2023-05-18
Feat: Sarama Kafka Integration

mirackara:
A integration for the [sarama](https://github.com/Shopify/sarama) package that instruments Kafka


v3/integrations/nrsarama/consumer.go
```diff
@@ -0,0 +1,131 @@
...
+	"context"
+	"net/http"
+
+	"github.com/Shopify/sarama"
+	"github.com/newrelic/go-agent/v3/internal"
+
+	"github.com/newrelic/go-agent/v3/newrelic"
+)
+
+func init() { internal.TrackUsage("integration", "messagebroker", "saramaconsumer") }
```
iamemilio, line 13:
```suggestion
func init() { internal.TrackUsage("integration", "messagebroker", "saramakhafka") }
```

I would suggest making this generic to this package because of the way the init() function works in go. It might be worth testing that it is working the way you think it is, but from what I understand about the init function in Go, all the inits for a `package` are executed the first time code from that package is imported and executed. That would probably mean in this case that even if you just had the consumer, we would still register the producer too. Is there a use case where you would use just the producer or consumer?

mirackara: There could definitely be a use case where a user could only be using the producer/consumer. If for example a kafka producer is created in a different language/package you can still use sarama's consumer to capture the messages sent. With that being said, I think I might've misunderstood the init() function. It seems like having one single unified metric across both is the correct implementation.

iamemilio (approved):

PR #724, opened 2023-06-01
oops release 3.22.0

iamemilio:
oopsie

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,26 @@
+## 3.22.0
+
+ ### Added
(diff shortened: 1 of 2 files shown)


nr-swilloughby (commented):
looks good, but if we already tagged 3.22.0 we can't retag it to a different commit, we need to release 3.22.1.

nr-swilloughby (approved):
I'm going to merge this and then follow it with a version bump and then release.

nr-swilloughby (approved):
I'm going to merge this and then follow it with a version bump and then release.

PR #727, opened 2023-06-07
Disable IAST if HighSecurity is enabled

aayush-ap:
Disabling IAST if HighSecurity is enabled because in HighSecurity mode request parameters can't be collected and without request parameters, IAST won't be able to work properly.

reference: https://docs.newrelic.com/docs/apm/agents/go-agent/configuration/go-agent-configuration/#high_security

Disabling IAST if HighSecurity is enabled


nr-swilloughby (changes requested):
Looks good but can you sign the contributor license agreement so that check can be cleared?
Thanks

aayush-ap:
> Looks good but can you sign the contributor license agreement so that check can be cleared? Thanks

Sure

nr-swilloughby (approved):

PR #729, opened 2023-06-12
Update handling for complete disabling of security agent.

aayush-ap:
- To completely disable security set NEW_RELIC_SECURITY_AGENT_ENABLED env to false. If the env is set to false, the security module and hooks are not loaded.
- Update default config YAML in README.md
- Updated go.mod in nrsecurityagent to point to the latest csec-go-agent version (v0.2.1)

### DIFF
--- a/v3/integrations/nrsecurityagent/nrsecurityagent.go
+++ b/v3/integrations/nrsecurityagent/nrsecurityagent.go
@@ -33,6 +33,17 @@ func defaultSecurityConfig() SecurityConfig {
 	return cfg
 }
 
+// To completely disable security set NEW_RELIC_SECURITY_AGENT_ENABLED env to false.
+// If env is set to false,the security module is not loaded
+func isSecurityAgentEnabled() bool {
+	if env := os.Getenv("NEW_RELIC_SECURITY_AGENT_ENABLED"); env != "" {
+		if b, err := strconv.ParseBool("false"); nil == err {
+			return b
+		}
+	}
+	return true
+}
+
 // InitSecurityAgent initializes the nrsecurityagent integration package from user-supplied
 // configuration values.
 func InitSecurityAgent(app *newrelic.Application, opts ...ConfigOption) error {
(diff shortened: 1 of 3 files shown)


v3/integrations/nrsecurityagent/nrsecurityagent.go
```diff
@@ -33,6 +33,17 @@ func defaultSecurityConfig() SecurityConfig {
 	return cfg
 }
 
+// To completely disable security set NEW_RELIC_SECURITY_AGENT_ENABLED env to false.
+// If env is set to false,the security module is not loaded
```
nr-swilloughby, line 37: This seems to be breaking with how configuration works for the agent and the nrsecurityagent integration, in that this keys exclusively on an environment variable while the rest of the configuration is done via environment variable, code, and/or config file. Instead of checking separately for the environment variable here, this code should check with the integration configuration.

aayush-ap: To disable the security agent we have two config flags.

```
c.Security.Enabled           // This is used to disable security agent functionality (hooks are applied but no event will be generated)
c.Security.Agent.Enabled // This is used to completely disable security. (security module and hooks are not loaded)

```
In csec-go-agent we apply hooks at the time of package initializing. At that time we don't have a config object that's why to completely disable security set `NEW_RELIC_SECURITY_AGENT_ENABLED` env to false.

v3/integrations/nrsecurityagent/nrsecurityagent.go
```diff
@@ -53,8 +64,7 @@ func InitSecurityAgent(app *newrelic.Application, opts ...ConfigOption) error {
 	if !isValid {
 		return fmt.Errorf("Newrelic application value cannot be read; did you call newrelic.NewApplication?")
 	}
-
-	if !appConfig.HighSecurity {
+	if !appConfig.HighSecurity && isSecurityAgentEnabled() {
```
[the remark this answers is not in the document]
aayush-ap: `c.Security.Enabled` flag is not used to completely disable security. This is used to disable security agent functionality (hooks will be applied but no event will be generated)

nr-swilloughby (approved):

Issue #743, opened 2023-06-23
No link between log message and application


EugeneBalybin:
## Description

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your New Relic account or API keys! )
Got the following setup:
Go application with newrelic/go-agent v3.20.3
```
newRelic, err := newrelic.NewApplication(
	newrelic.ConfigAppName(*cfg.NewRelic.AppName),
	newrelic.ConfigLicense(*cfg.NewRelic.LicenseKey),
	newrelic.ConfigAppLogForwardingEnabled(false),
	newrelic.ConfigDistributedTracerEnabled(true),
	newrelic.ConfigAppLogDecoratingEnabled(true),
);
writer := zerologWriter.New(os.Stdout, newRelic)
logger := zerolog.New(writer)
logger.Info().Msg("test message")
```
`nri-bundle` running for log forwarding to NewRelic

With the code above I'm getting the log message to stdout:
`{"level":"info","message":"test message"} NR-LINKING|xxxYYYzzzQQQ|HostName|AppName|`

`nri-bundle` gets the message, sends it to the NewRelic, and I can find it in the New Relic logs, but they are in the global logs section, not in the APM -> Logs. 

In the New Relic UI there is no "NR-LINKING" part, just json which is properly decoded into a message fields. And among the fiends there are no entity.name or entity.guid fields 

BUT I can see log metrics on the APM summary page 

## Expected Behavior

entity.name and entity.guid are present and logs discoverable in the APM section

EugeneBalybin:
So I played around and found that NR-LINKING part should not omit pipes when outside the transaction
Like this
`{"level":"info","message":"test message"} NR-LINKING| xxxYYYzzzQQQ|HostName|||AppName|`

EugeneBalybin:
https://github.com/newrelic/go-agent/pull/744

iamemilio:
The next go agent release will include this.

PR #744, opened 2023-06-23
Fix NR-LINKING 

EugeneBalybin:
## Links
Fixes issue: https://github.com/newrelic/go-agent/issues/743

## Details
Pipes should be preserved for newrelic backend


iamemilio (approved):
Thanks for contributing

Issue #748, opened 2023-06-27
3.23 Regression: parallel NewGoroutine calls result in data races


adomaskizogian:
We have observed that running application with race detector enabled (`-race`) it detects that parallel calls to `newrelic.FromContext(ctx).NewGoroutine()` results in data races.

No issues with `3.22`

sample code that calls `NewGoroutine`

```golang
func NewNewrelicGoroutineTransaction(ctx context.Context) (context.Context, *newrelic.Transaction){
   txn := newrelic.FromContext(ctx)
   if txn != nil {
    gTxn := txn.NewGoroutine()  
    ctx = newrelic.NewContext(ctx, gTxn)
    return  ctx, gTxn
  }
  return ctx, nil
}
```

```
WARNING: DATA RACE
Write at 0x00c0010cc6d8 by goroutine 2026:
  github.com/newrelic/go-agent/v3/newrelic.(*Transaction).NewGoroutine()
      /home/adomas.kizogian/go/pkg/mod/github.com/newrelic/go-agent/v3@v3.23.0/newrelic/transaction.go:502 +0x1db
```

rittneje:
@iamemilio This bug prevents us from upgrading from 3.22.1. Can you please resolve it? Also, this library ought to run its unit tests with the race detector, in order to detect such bugs during development.

nr-swilloughby:
We're working to release a fix for this ASAP.

nr-swilloughby:
Fix released

adomaskizogian:
Thank you guys!

rittneje:
@nr-swilloughby In the future it would be nice to include fixes like this in the release notes.

nr-swilloughby:
@rittneje sorry about that. The release notes mentioned the fix that was made, which happened to be the cause of both a performance issue and the race condition, but neglected to actually say that it fixed the race condition as well. I'll make a note about that in the next release.

PR #754, opened 2023-07-11
Fix for data race with NewGoroutine

aayush-ap:
- added fix for data race with NewGoroutine.
https://github.com/newrelic/go-agent/issues/748

### DIFF
--- a/v3/newrelic/internal_txn.go
+++ b/v3/newrelic/internal_txn.go
@@ -1363,3 +1363,17 @@ func (txn *txn) IsSampled() bool {
 
 	return txn.lazilyCalculateSampled()
 }
+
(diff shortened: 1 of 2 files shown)


nr-swilloughby (approved):
Approved, but I think before we release this we'll add a guard to the calls to `getCsecData` and `setCsecData` to prevent them if the security agent is not enabled. Although in that case, `secureAgent.SendEvent` just returns `nil` and will likely be inlined by the compiler, the evaluation of the new function calls will result in an extra transaction-wide lock every time a segment is started. We'll add a check so this only happens if csec data will actually be used.

nr-swilloughby:
(...and verify that the implementation of the above fully resolves the race condition issue)

PR #774, opened 2023-08-25
(WIP) FastHTTP Integration

mirackara:


iamemilio:
If you're going to wrap the client, then you should provide a wrapped function for each method of the client object, even if they are self referential https://pkg.go.dev/github.com/valyala/fasthttp#Client

v3/integrations/nrfasthttp/nrfasthttp_client.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/valyala/fasthttp"
+)
+
+type FastHTTPClient interface {
+	Do(req *fasthttp.Request, resp *fasthttp.Response) error
+}
+
+func Do(client FastHTTPClient, txn *newrelic.Transaction, req *fasthttp.Request, res *fasthttp.Response) error {
+	seg := txn.StartSegment("fasthttp-do")
+	err := client.Do(req, res)
```
iamemilio, line 14: Should we inject DT headers into the request? This can get tricky. We may also want to consider filling out more info in the transaction. Like whether we are categorizing this request as a web/non-web request.

v3/integrations/nrfasthttp/example/main.go
```diff
@@ -0,0 +1,56 @@
...
+	"time"
+
+	"github.com/newrelic/go-agent/v3/integrations/nrfasthttp"
+
+	newrelic "github.com/newrelic/go-agent/v3/newrelic"
+
+	"github.com/valyala/fasthttp"
+)
+
+func main() {
```
iamemilio, line 15: I would like to see a client example too. They are different products, and we should show customers how to use both.

v3/integrations/nrfasthttp/nrfasthttp_client.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	newrelic "github.com/newrelic/go-agent/v3/newrelic"
+	"github.com/valyala/fasthttp"
+)
+
+type FastHTTPClient interface {
+	Do(req *fasthttp.Request, resp *fasthttp.Response) error
+}
+
+func Do(client FastHTTPClient, txn *newrelic.Transaction, req *fasthttp.Request, res *fasthttp.Response) error {
```
iamemilio, line 12: We should provide wrappers for all the methods of the Client object. Otherwise, customers will not have a way to wrap them.

v3/integrations/nrfasthttp/nrfasthttp_client.go
```diff
@@ -0,0 +1,21 @@
+package nrfasthttp
+
+import (
+	newrelic "github.com/newrelic/go-agent/v3/newrelic"
+	"github.com/valyala/fasthttp"
+)
+
+type FastHTTPClient interface {
```
iamemilio, line 8: maybe providing a wrapped object like you do for request might be a better design pattern here.

v3/integrations/nrfasthttp/nrfasthttp_request.go
```diff
@@ -0,0 +1,75 @@
...
+		segRequest := txn.StartSegment("fasthttp-set-request")
+		// Set transaction attributes
+		txn.AddAttribute("method", string(ctx.Method()))
+		txn.AddAttribute("path", string(ctx.Path()))
+		// convert fasthttp request to http request
+		r := &http.Request{}
+		fasthttpadaptor.ConvertRequest(ctx, r, true)
+
+		txn.SetWebRequestHTTP(r)
+		txn.InsertDistributedTraceHeaders(r.Header)
```
iamemilio, line 65:
```suggestion
```
We should not add DT headers here. https://github.com/newrelic/go-agent/blob/master/v3/newrelic/instrumentation.go#L35

v3/integrations/nrfasthttp/nrfasthttp_request.go
```diff
@@ -0,0 +1,75 @@
...
+	txn := ctx.UserValue("transaction")
+
+	if txn == nil {
+		return nil
+	}
+
+	return txn.(*newrelic.Transaction)
+}
+
+func NRHandler(app *newrelic.Application, original fasthttp.RequestHandler) fasthttp.RequestHandler {
```
iamemilio, line 44: We want this to have the same behaviors as the go agent does. In oder to do that, this needs to be in the v3/newrelic directory. I think its worth considering making this part of our core instrumentation https://github.com/newrelic/go-agent/blob/master/v3/newrelic/instrumentation.go#L35

v3/newrelic/internal_17_test.go
```diff
@@ -18,6 +19,37 @@ func myErrorHandler(w http.ResponseWriter, req *http.Request) {
...
+
+	if wrappedHandler == nil {
+		t.Error("Error when creating a wrapped handler")
+	}
+	ctx := &fasthttp.RequestCtx{}
+	ctx.Request.Header.SetMethod("GET")
+	ctx.Request.SetRequestURI("/hello")
+	wrappedHandler(ctx)
+
+	app.ExpectMetrics(t, []internal.WantMetric{
```
iamemilio, line 40: I noticed the wrapHandle test also creates and expects an error. Can we do that for the tests for fasthttp as well?

mirackara: there's an example of noticing an error in the fasthttp-server example :) I don't think it should be too difficult to add that into the internal testing as well.

iamemilio:
Over all, this looks really good :)

Everything seems to be working as expected and I like the user experience much better of this version

iamemilio (approved):

v3/newrelic/segments.go
```diff
@@ -337,6 +340,30 @@ func StartExternalSegment(txn *Transaction, request *http.Request) *ExternalSegm
...
+		txn = transactionFromRequestContextFastHTTP(ctx)
+	}
+	request := &http.Request{}
+
+	fasthttpadaptor.ConvertRequest(ctx, request, true)
+	s := &ExternalSegment{
+		StartTime: txn.StartSegmentNow(),
+		Request:   request,
+	}
+	s.secureAgentEvent = secureAgent.SendEvent("OUTBOUND", request)
```
nr-swilloughby, line 354: This statement should be inside an `if IsSecurityAgentPresent(){ ... }` conditional

v3/newrelic/segments.go
```diff
@@ -337,6 +340,30 @@ func StartExternalSegment(txn *Transaction, request *http.Request) *ExternalSegm
...
+	if request != nil && request.Header != nil {
+		for key, values := range s.outboundHeaders() {
+			for _, value := range values {
+				request.Header.Set(key, value)
+			}
+		}
+		secureAgent.DistributedTraceHeaders(request, s.secureAgentEvent)
```
nr-swilloughby, line 361: This statement should be inside an `if IsSecurityAgentPresent(){ ... }` conditional

nr-swilloughby (approved):
Looks great to me now.

v3/examples/client-fasthttp/main.go
```diff
@@ -0,0 +1,62 @@
...
+	defer fasthttp.ReleaseRequest(req)
+	defer fasthttp.ReleaseResponse(resp)
+
+	req.SetRequestURI("http://localhost:8080/hello")
+	req.Header.SetMethod("GET")
+
+	ctx := &fasthttp.RequestCtx{}
+	seg := newrelic.StartExternalSegmentFastHTTP(txn, ctx)
+	defer seg.End()
+
```
aayush-ap, line 26: `ctx` doesn't contain any outbound req attributes. it's just an empty instance of RequestCtx.
  We would require outbound req URI and header for generating trace events.
   @mirackara

v3/newrelic/segments.go
```diff
@@ -337,6 +340,36 @@ func StartExternalSegment(txn *Transaction, request *http.Request) *ExternalSegm
...
+	fasthttpadaptor.ConvertRequest(ctx, request, true)
+	s := &ExternalSegment{
+		StartTime: txn.StartSegmentNow(),
+		Request:   request,
+	}
+	if IsSecurityAgentPresent() {
+		s.secureAgentEvent = secureAgent.SendEvent("OUTBOUND", request)
+	}
+
+	if request != nil && request.Header != nil {
```
aayush-ap, line 358: `s.outboundHeaders` and `DistributedTraceHeaders` should be added to the original request context before executing the request.

aayush-ap, line 358: `s.outboundHeaders` and `DistributedTraceHeaders` should be added to the original request context before executing the request.

nr-swilloughby:
I'll open a bug fix for this.

v3/newrelic/segments.go
mirackara: Hi @aayush-ap

Looking at this, the StartExternalSegmentFastHTTP function is identical to the StartExternalSegment function. The only difference being is that we convert to a http request. Im a bit confused at this comment because we are adding `s.outboundHeaders` and `DistributedTraceHeaders` before returning the request context and executing the request.


 Unless you mean calling `s.secureAgentEvent = secureAgent.SendEvent("OUTBOUND", request)` *after* we add the two headers?

cc @nr-swilloughby

v3/examples/client-fasthttp/main.go
mirackara: Is this still the case if this is on the client side? In the regular [client example](https://github.com/newrelic/go-agent/blob/d0a46d00dd180231a7dd41c973c78b32bb56c70d/v3/examples/client/main.go#L15C2-L15C2) we do the same thing as well.

aayush-ap: @mirackara In the regular client example we are passing HTTP request object as an attribute of  **[StartExternalSegment](https://github.com/newrelic/go-agent/blob/master/v3/examples/client/main.go#L21)**  which contain outbound req parameters like URL and all.

but in the case of fastHttp you are passing fast HTTP RequestCtx object as an attribute of  [StartExternalSegmentFastHTTP](https://github.com/newrelic/go-agent/blob/master/v3/examples/client/main.go#L21)
ctx doesn't contain any outbound req attributes. it's just an empty instance of RequestCtx.


example:

```
req := fasthttp.AcquireRequest()
  .....
  ......
 req.SetRequestURI("https://www.google.com/")
 ctx := &fasthttp.RequestCtx{}
 seg := newrelic.StartExternalSegmentFastHTTP(txn, ctx)
 defer seg.End()

 err := fasthttp.Do(req, resp)

```

- Expected :
  {
  "name":"External/[www.google.com/https](http://www.google.com/https)",
  "scope":"WebTransaction/Go/GET /case1"
 }
 
 - Actual :
  {
    "name":"External/unknown/http/GET",
    "scope":"WebTransaction/Go/GET /doRequest"
 },

v3/newrelic/segments.go
aayush-ap: Yes,  you are converting `ctx` fasthttp.RequestCtx to http request and then adding all the required headers.
But that does not reflect the original request object that is used to call an outbound request.

v3/newrelic/instrumentation.go
```diff
@@ -5,18 +5,41 @@ package newrelic
...
+	"github.com/valyala/fasthttp"
+	"github.com/valyala/fasthttp/fasthttpadaptor"
 )
 
+type fasthttpWrapperResponse struct {
+	ctx *fasthttp.RequestCtx
+}
+
+func (rw fasthttpWrapperResponse) Header() http.Header {
+	hdrs := http.Header{}
```
aayush-ap, line 18: Not able to access inbound request response with  `fasthttpWrapperResponse`  wrapper

v3/examples/client-fasthttp/main.go
aayush-ap: Please check the following fix for outbound request

https://github.com/k2io/go-agent-fork/commit/15a72c8992aa1767f7b2596ed84d07307a1d23d3

v3/newrelic/segments.go
aayush-ap: Please check the following fix for outbound request

https://github.com/k2io/go-agent-fork/commit/15a72c8992aa1767f7b2596ed84d07307a1d23d3

PR #775, opened 2023-08-28
Minor fixes for security agent.

aayush-ap:
- minor fix for complete security disable flag
- Update security agent version to the latest (v0.4.0)


MadScienceZone (approved):

nr-swilloughby (approved):
Let's try approving this from the correct account this time.

PR #778, opened 2023-09-11
chore: add logs-in-context example with logrus

Julien4218:
## Details

This is a variation of the existing example for http-server, adding the specific steps to enable logs-in-context with distributed tracing.

image


iamemilio:
rather than contribute this to v3/examples I think this would be a better fit for v3/integrations/logcontext-v2/nrlogrus/examples. Would you mind moving the file there? Thanks for contributing though, we are always happy to have more examples.

Julien4218:
> rather than contribute this to v3/examples I think this would be a better fit for v3/integrations/logcontext-v2/nrlogrus/examples. Would you mind moving the file there? Thanks for contributing though, we are always happy to have more examples.

Sounds good, done.

iamemilio (approved):

iamemilio:
Once you sign the CLA we can merge this.

Julien4218:
> Once you sign the CLA we can merge this.

Done, thank you.

Issue #639, opened 2023-02-04
Delete old v2 code


rittneje:
Considering what happened in #636, and the fact that v2 is now [completed unsupported](https://docs.newrelic.com/docs/apm/agents/go-agent/get-started/go-agent-eol-policy/), all the old v2 code should be deleted to avoid confusion.

That is:
1. all the .go files at the root of the repo
2. the [_integrations](https://github.com/newrelic/go-agent/tree/master/_integrations) folder
3. the [examples](https://github.com/newrelic/go-agent/tree/master/examples) folder
4. the [internal](https://github.com/newrelic/go-agent/tree/master/internal) folder

nr-swilloughby:
Agreed, this is already on our roadmap to do.

PR #798, opened 2023-10-17
removed v2 code

nr-swilloughby:
This removes the obsolete V2 agent code from the product. Only the currently-supported agent version (v3) now will be here going forward.


nr-swilloughby:
Fixes [Issue 639](https://github.com/newrelic/go-agent/issues/639)

iamemilio (approved):

Issue #795, opened 2023-10-04
Consumes a lot of memory during large body size requests


juri-t:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

When we send a request with a large body size, it consumes a lot of memory because NewRelic is putting all the bodies in memory with io.ReadAll. Depending on the server's memory, this can bring down the server with an Out of Memory error.

https://github.com/newrelic/go-agent/blob/master/v3/newrelic/transaction.go#L255

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your New Relic account or API keys! )

## Steps to Reproduce

```golang
package main

import (
	"github.com/labstack/echo/v4"
	"github.com/newrelic/go-agent/v3/integrations/nrecho-v4"
	"github.com/newrelic/go-agent/v3/newrelic"
	"io"
)

func main() {
	e := echo.New()

	app, _ := newrelic.NewApplication(
		newrelic.ConfigAppName("App Name"),
		newrelic.ConfigLicense("Licence Key"),
		func(cfg *newrelic.Config) {
			cfg.Enabled = true
		},
	)

	e.POST("/", func(ec echo.Context) error {
		io.Copy(io.Discard, ec.Request().Body)
		return ec.NoContent(200)
	}, nrecho.Middleware(app))
	e.Start(":3000")
}
```

POST requests for large files created by mkfile, etc., consume a lot of memory.

[NOTE]: # ( Please be as specific as possible. )

## Expected Behavior
Does not consume large amounts of memory.

[NOTE]: # ( Tell us what you expected to happen. )

## NR Diag results

[NOTE]: # ( Provide any other relevant log data. )

## Your Environment

[TIP]:  # ( Include as many relevant details about your environment as possible including the running version of New Relic software and any relevant configurations. )

## Reproduction case

[TIP]: # ( Link a sample application that demonstrates the issue. )

## Additional context

[TIP]:  # ( Add any other context about the problem here. )

mirackara:
Hello @juri-t, thanks for the report! We are currently investigating this.

iamemilio:
This is released in the 3.28.0 version of the go agent!

PR #806, opened 2023-10-23
Fix for out of memory error with request body

aayush-ap:
## Fix for out-of-memory error with request body
 - https://github.com/newrelic/go-agent/issues/795
 - Added new implementation to add a limit on request body size.
  - To configure body_limit on read request body use `nrsecurityagent.ConfigSecurityRequestBodyLimit` method or `NEW_RELIC_SECURITY_REQUEST_BODY_LIMIT`  env. By default, this is "300kb"
- Example :
export NEW_RELIC_SECURITY_REQUEST_BODY_LIMIT=100
- When the security agent is not active, the APM agent does not access the request body. 
- This change will required min v0.5.0 version of csec-go-agent


v3/newrelic/transaction.go
```diff
@@ -633,7 +626,7 @@ func (webrequest WebRequest) GetHost() string {
 	return webrequest.Host
 }
 
-func (webrequest WebRequest) GetBody() []byte {
+func (webrequest WebRequest) GetBody() interface{} {
```
mirackara, line 629: Replacing `interface{}` with `any` type here would be best practice

mirackara:
@aayush-ap It looks like this change also breaks the `nrmicro` and `nrgrpc` integrations (You can see more details in the GHA test suite for those two tests respectively. Once these are fixed, this looks good to merge :)

aayush-ap:
@mirackara 
Please verify with latest changes

mirackara (approved):
LGTM! Thank you

v3/newrelic/transaction.go
```diff
@@ -633,7 +626,7 @@ func (webrequest WebRequest) GetHost() string {
 	return webrequest.Host
 }
 
-func (webrequest WebRequest) GetBody() []byte {
+func (webrequest WebRequest) GetBody() any {
```
mirackara, line 629: Hello @aayush-ap, is there a specific reason why this is an any type? Is this function to be used only for the security agent? If there is a specific type we want to make sure any external usage has an expected return value.

aayush-ap: yes,getBody is only used by the cses-go-agent(security agent).
https://github.com/newrelic/csec-go-agent/blob/main/security_intercept/intercept.go#L748

aayush-ap: For now this is not possible because csec-go-agent has already released we can make this change in the next release of both agents.
According to me, this will not make an issue because GetBody only used by the csec agent.

mirackara: Alright. I think we're going to put this on ice for now until the changes to the csec-go-agent are in place. We just want to make sure that any public facing API's remain the same so customer who *may be* using the function in their code bases aren't affected by this change

mirackara (approved):

Issue #807, opened 2023-10-25
Remove fasthttp dependency from main module


rittneje:
## Summary

As of v3.25.0 the github.com/newrelic/go-agent/v3 module has a direct dependency on github.com/valyala/fasthttp.

We have to carefully review third-party libraries. This one in particular is rather concerning due to its use of `unsafe`.

Judging from the code, it looks like the main module doesn't even need this dependency. It's just so that anyone who happens to use `fasthttp` can instrument their code. But this should have been isolated to an extension like everything else.

## Desired Behavior

1. The public API of the main module should not include anything outside the Go standard library.
2. The main module should not include dependencies other than to support its public API and its backing implementation.

## Possible Solution

It's a breaking change, so it's tricky. Either just make a breaking change in a minor release and direct people to use the new extension, or release a new major version.

iamemilio:
scheduled for the next release

PR #808, opened 2023-10-25
move fasthttp out of core library, and into integration package

iamemilio:
## Links

Possible solution to #807


v3/integrations/nrfasthttp/segment.go
```diff
@@ -0,0 +1,56 @@
...
+		StartTime: txn.StartSegmentNow(),
+		Request:   request,
+	}
+
+	if newrelic.IsSecurityAgentPresent() {
+		secureAgentEvent = newrelic.GetSecurityAgentInterface().SendEvent("OUTBOUND", request)
+		s.SetSecureAgentEvent(secureAgentEvent)
+	}
+
+	if request != nil && request.Header != nil {
```
mirackara, line 30: Since we're moving this code, it may be a good idea to also patch the [small issues related to the security agent](https://github.com/k2io/go-agent-fork/commit/15a72c8992aa1767f7b2596ed84d07307a1d23d3).

v3/integrations/nrfasthttp/examples/server-fasthttp/main.go
mirackara, line 1: Even though fasthttp is in it's own integration folder now, might it be a good idea to still keep the server example in the v3/examples folder?

iamemilio: I was thinking that, but it depends on the v3/go.mod file which would either require the Fasthttp dependency to be in it, or cause the tests to fail

v3/integrations/nrfasthttp/segment.go
iamemilio: this should address https://new-relic.atlassian.net/browse/NR-169080 now

iamemilio:
This needs some revision, we can not copy the request object

iamemilio:
The more I look at this, the more I realize there may not be a good way to capture a Fasthttp.request. It will always throw an error because of the copy operation, and there is not good way to convert it to an object we can read and use.

nr-swilloughby (approved):

nr-swilloughby (changes requested):
The thing about copying the request was bugging me too... let's hold off just yet.

nr-swilloughby (approved):
Much better! yay!

v3/go.mod
```diff
@@ -4,7 +4,8 @@ go 1.19
 
 require (
 	github.com/golang/protobuf v1.5.3
-	google.golang.org/grpc v1.54.0
+	github.com/valyala/fasthttp v1.49.0
```
mirackara, line 7: we should remove this from the v3 go.mod

mirackara (approved):

PR #826, opened 2023-11-23
feat instrument pgx CopyFrom

### LINKED ISSUE #825: Add support to CopyFrom method in pgx integration
[NOTE]: # ( ^^ Provide a general summary of the request in the title above. ^^ )

## Summary

Hi, currently we are using pgx methods `CopyFrom`, that uses postgresql copy protocol, on ours bulk inserts.
Pgx provides an interface from this method on `Tracer` interface but it's not implemented on pgx v5 integration.
I have found a open telemetry integration that implements support to CopyFrom [link](https://github.com/exaring/otelpgx/blob/main/tracer.go#L170)

[NOTE]: # ( Provide a brief overview of what the new feature is all about. )

## Desired Behaviour

[NOTE]: # ( Tell us how the new feature should work. Be specific. )
[TIP]:  # ( Do NOT give us access or passwords to your New Relic account or API keys! )

When any `CopyFrom` methods is called should appear a span on new relic request trace.

## Possible Solution

[NOTE]: # ( Not required. Suggest how to implement the addition or change. )

I think we can base on otel implementation. If it's needed, I can provide a PR for this feature

## Additional context

[TIP]:  # ( Why does this feature matter to you? What unique circumstances do you have? )

vinicius-batista:
## Links
Closes #825

## Details

The implementation and tests was simple but I have a problem running tests when snapshot was already created.
The current installed version from `pgsnap` fails when read `CopyData` from snapshot.
At pgsnap's latest version CopyData and other messages were implemented but pgproto is failing to unmarshal `CopyInResponse` message.

I don't know how to proceed  with this implementation as some tests are failing due to external dependencies.

### DIFF
--- a/v3/integrations/nrpgx5/nrpgx5.go
+++ b/v3/integrations/nrpgx5/nrpgx5.go
@@ -16,36 +16,36 @@
 //
 // For example:
 //
-//    import (
-//       "github.com/jackc/pgx/v5"
-//       "github.com/newrelic/go-agent/v3/integrations/nrpgx5"
-//       "github.com/newrelic/go-agent/v3/newrelic"
-//    )
+//	import (
+//	   "github.com/jackc/pgx/v5"
+//	   "github.com/newrelic/go-agent/v3/integrations/nrpgx5"
+//	   "github.com/newrelic/go-agent/v3/newrelic"
+//	)
 //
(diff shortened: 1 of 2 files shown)


vinicius-batista:
@iamemilio

PR #835, opened 2023-12-11
Text language updates

mirackara:


iamemilio:
if we change the docs to https for how to connect to the example server, we should set the example server up over https too

iamemilio (approved):

nr-swilloughby (approved):

PR #854, opened 2024-02-16
conditional instrumentation and profiling options for server example

nr-swilloughby:
I'm just going to park this here to discuss whether we want to include this in the agent.

This adds some compile-time flags for the server example app so you can easily build it with and without the agent, and with and without profiling. The default is the same as always (no profiling, yes agent). This may be of use for doing speed comparisons.

On the other hand, it does add some clutter to the code in the example app. The question is whether the benefit is worth the clutter.


nr-swilloughby:
In the interest of keeping the code examples as clean and straightforward as possible, let's hold off on this for now.

PR #860, opened 2024-02-27
OpenAI Integration (WIP)

mirackara:
Contains instrumentation for the following functions from the [go-openai](https://github.com/sashabaranov/go-openai) library

- `CreateChatCompletion`
- `CreateChatCompletionStream`
- `CreateEmbeddings`

- Adds new Event type called `RecordLlmFeedbackEvent` which can be used to record feedback for any given chat completion response


v3/integrations/nropenai/nropenai.go
```diff
@@ -8,18 +8,23 @@ import (
...
 	"strings"
+	"time"
 
 	"github.com/google/uuid"
+	"github.com/newrelic/go-agent/v3/internal"
 	"github.com/newrelic/go-agent/v3/newrelic"
 	"github.com/sashabaranov/go-openai"
 )
 
+func init() { internal.TrackUsage("Go", "ML", "OpenAI", "1.20.2") }
```
nr-swilloughby, line 19: If I understand the spec correctly, this should be reporting the version of the library the customer actually linked with their code, not a hardcoded value.

nr-swilloughby (approved):

PR #872, opened 2024-03-21
Bedrock Integration

nr-swilloughby:
I pulled back the stream support momentarily as it still needs a little more work. That'll be readded shortly.
This implements AWS Bedrock instrumentation.


nr-swilloughby:
need to redo this, it includes unwanted changes to the integrations.

PR #874, opened 2024-03-21
bedrock tweaks after testing

nr-swilloughby:
This PR includes a few post-testing adjustments to the bedrock implementation.


nr-swilloughby:
As requested, this also removes the size limits for custom event attributes only if they are LLM custom events. Also enables the sending of LLM custom events regardless of whether high security mode or custom insights are enabled.

mirackara (approved):

PR #881, opened 2024-03-26
feat: slog: automatically reference transaction from context

adomaskizogian:
I would like to introduce a handler that takes new relic transaction data from context automatically.

It is quite a common pattern to inject the base slog logger instance to some piece of code only so to pass it to `nrslog.WithContext` to receive a handler that captures logs in context. I find this api to be a bit leaky and introducing noise. Especially that slog already has methods which make context available for handlers.

With introduction of this handler one now can set a slog instance relying on a NRHandler as default one and call it from anywhere, ensuring that log data with be passed to go-agent, as well as if the context contains transaction it will be linked. This is quite the case for http services which rely on `newrelic.WrapHandle`. This also decouples non infra/non instrumentation code from go-agent and allows to write code using stdlib.

an example:

on main.go
```go
app, _ := newrelic.NewApplication(
  newrelic.ConfigAppName("foo"),
  newrelic.ConfigLicense("bar"),
)

instrumentedHandler := nrslog.WithTransactionFromContext(nrslog.TextHandler(app, os.Stderr, &slog.HandlerOptions{}))
slog.SetDefault(slog.New(instrumentedHandler))
```

on service.go

```go
func DoSomeWork(ctx context.Context, app *newrelic.Application){
  txn := app.StartTransaction("baz")
  defer txn.End()

  ctx = newrelic.NewContext(ctx, txn)

  slog.InfoContext(ctx, "qux")  
}
```

All feedback is welcome. The implementation itself was hacked out pretty quick. I am open for suggestions for taking a different approach such as re-implementing the handler. At the moment this new feature is implemented using a decorator which disables setting the transactions externally and only allows passing transaction via context.

I hope this will serve as a good starting point for a discussion. Ideally the resulting implementation will not require to change the example above and the usage remains the same.

btw, not sure if the tests will pass as I found no way to run tests on local env.
also, slack link is broken on CONTRIBUTING.md

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,6 +30,7 @@ jobs:
           - dirs: v3/integrations/logcontext-v2/nrlogrus
           - dirs: v3/integrations/logcontext-v2/nrzerolog
           - dirs: v3/integrations/logcontext-v2/nrzap
+          - dirs: v3/integrations/logcontext-v2/nrslog
           - dirs: v3/integrations/logcontext-v2/nrwriter
           - dirs: v3/integrations/logcontext-v2/zerologWriter
           - dirs: v3/integrations/logcontext-v2/logWriter
--- a/v3/integrations/logcontext-v2/nrslog/handler.go
+++ b/v3/integrations/logcontext-v2/nrslog/handler.go
@@ -215,18 +215,9 @@ func WithTransactionFromContext(handler NRHandler) TransactionFromContextHandler
 //   - If a group has no Attrs (even if it has a non-empty key),
 //     ignore it.
 func (h TransactionFromContextHandler) Handle(ctx context.Context, record slog.Record) error {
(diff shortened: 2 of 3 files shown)


iamemilio:
@adomaskizogian thanks for the patch, is this ready for me to take a look? If not, @ me when it is.

adomaskizogian:
> @adomaskizogian thanks for the patch, is this ready for me to take a look? If not, @ me when it is.

it would be really helpful if you could share the details how to run tests of the nrslog package on local dev

iamemilio:
I do most of my development in vscode, so I use the built in test tools to run targeted unit tests. Its much easier than learning the test CLI. Good spot on the CI missing the test though.

iamemilio:
You should not need anything special to run the tests though, and from what I can tell the test you created is structured correctly. I authorized the CI to run for you.

iamemilio (approved):
lgtm

PR #899, opened 2024-04-11
Fix for correct struct field name for echo version v3.2.2+

aayush-ap:
Fix for correct struct field name for echo version v3.2.2+

See [here](https://pkg.go.dev/gopkg.in/labstack/echo.v3@v3.2.2#Route) for documentation related to echo version v3.2.2 and above.
For echo version up to v3.2.1:

```
type Route struct {
	Method  string `json:"method"`
	Path    string `json:"path"`
	Name  string `json:"name"`
}
```

See [here](https://pkg.go.dev/gopkg.in/labstack/echo.v3@v3.2.1#Route) for documentation related to echo version v3.2.1.
```
type Route struct {
	Method  string `json:"method"`
	Path    string `json:"path"`
	Handler string `json:"handler"`
}
```

### DIFF
--- a/v3/integrations/nrecho-v3/nrecho.go
+++ b/v3/integrations/nrecho-v3/nrecho.go
@@ -35,6 +35,22 @@ func handlerPointer(handler echo.HandlerFunc) uintptr {
 	return reflect.ValueOf(handler).Pointer()
 }
 
+func handlerName(router interface{}) string {
+	val := reflect.ValueOf(router)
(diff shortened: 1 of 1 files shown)


iamemilio (approved):
This seems reasonable

nr-swilloughby:
Merged for testing. I'm updating the dependency to the csec library 1.2.0 before releasing.

PR #902, opened 2024-04-18
Zap Field Attributes Reporting

mirackara:
Adds support for Zap field attributes properly showing up on the New Relic Platform.


v3/newrelic/config.go
```diff
@@ -582,6 +582,10 @@ type ApplicationLogging struct {
 		// Toggles whether the agent enriches local logs printed to console so they can be sent to new relic for ingestion
 		Enabled bool
 	}
+	ZapLogger struct {
```
iamemilio, line 585: Could we add more info here about when it makes sense to do this: We want to enable this when your app collects fewer logs, or if your app can afford to compile the json during log collection, slowing down the execution of the line of code that will write the log. Otherwise, if your application collects logs at a high frequency or volume, or it can not afford the slowdown of marshaling objects before sending them to newrelic, we can marahal them asynchronously in the backend during harvests

iamemilio:
one nit pick, otherwise lgtm

PR #922, opened 2024-06-27
Changes Required by CSEC Agent

aayush-ap:
Added following chnages required by csec agent

* Added support to detect gRPC API endpoint.
* Added feature to detect route of an incoming request for all supported frameworks.
* Added support to send HTTP status code.


nr-swilloughby (approved):
provisionally approving; will merge and do more testing before releasing.

PR #932, opened 2024-07-05
Add log/slog support for agent logs

digitalcrab:
This change adds integration with standard `log/slog` package for agent logs.


iamemilio:
Thanks for contributing. Nice tests btw!

PR #935, opened 2024-07-17
Added Attribute Support for ZeroLog

mirackara:


iamemilio:
I think this will work with simple attributes like strings or numbers, but what about maps, or slices? I think that is a weakness of this library.

nr-swilloughby:
> I think this will work with simple attributes like strings or numbers, but what about maps, or slices? I think that is a weakness of this library in general. It needs to be able to parse a lot of complex objects, and the token that seperates them is configurable in zerolog. See my risk matrix item for more info. Before we spend more time here, I might recommend another re-write.

I agree. This is definitely a concern about the library.

PR #940, opened 2024-08-01
remove all extraneous version constraints

rittneje:
Fixes #939.

As per the README, Go 1.18+ is already required. Thus version constraints for 1.9 and 1.10 are unnecessary. (And these currently cause issues with 1.23rc2 due to https://github.com/golang/go/issues/68658.)

PR #948, opened 2024-08-15
fix(protobuf-deps): use latest grpc and protobuf

gunturaf:
## Links

https://github.com/newrelic/go-agent/pull/940#issuecomment-2291023624

## Details

v3.34.0 broke the world because:
1. It refers to the invalid protobuf version.
2. The generated protobuf+grpc files uses old versions of the protobuf+grpc deps.

This MR upgrades those both to the latest.
Also, as a side effect, this will upgrade the minimum Go version to use this library to Go 1.21.x
Finally, this MR also retracts v3.34.0


iamemilio:
This needs to be rebased on develop. Its bringing in a lot of extra commits.

gunturaf:
done @iamemilio 👍🏼

iamemilio (approved):

PR #952, opened 2024-08-21
Cloud Services Entity Relationship Changes

mirackara:


v3/integrations/nramqp/nramqp.go
```diff
@@ -33,13 +34,29 @@ func creatProducerSegment(exchange, key string) *newrelic.MessageProducerSegment
 	return &s
 }
 
+func GetHostAndPortFromURL(url string) (string, string) {
+	// url is of format amqp://user:password@host:port
+	// extract host from url after "@" symbol
+	parts := strings.Split(url, "@")
+	if len(parts) != 2 {
+		return "", ""
```
nr-swilloughby, line 42: I thought the user:password@ part of the uri was optional. Should we consider the case of "amqp://host:port" here instead of returning empty strings?

nr-swilloughby (changes requested):
just one question about uri format

nr-swilloughby (approved):

iamemilio (approved):

PR #956, opened 2024-08-28
Enhanced Security Features

aayush-ap:
Expanded security capabilities by adding support for Secure Cookie event reporting, providing detailed vulnerability information.
Enabled sharing of response headers with the CSEC-Security-Agent to extract cookie values from headers.


iamemilio (approved):
This is passing its tests, and seems uncontroversial.

PR #958, opened 2024-09-04
added mutex guards around appRun for code level metrics

nr-swilloughby:
Corrects access to the internal `app.app.run` structure by protecting that access with a mutex, in order to prevent race conditions.

### DIFF
--- a/v3/go.mod
+++ b/v3/go.mod
@@ -3,10 +3,9 @@ module github.com/newrelic/go-agent/v3
 go 1.20
 
 require (
-	github.com/golang/protobuf v1.5.3
 	google.golang.org/grpc v1.56.3
+	google.golang.org/protobuf v1.5.3
 )
-
 
 retract v3.22.0 // release process error corrected in v3.22.1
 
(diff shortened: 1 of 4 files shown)


iamemilio:
LGTM. Can we drop the go.mod changes from this pr though? That is handled in #948.

We're in a tricky spot because a lot is build on the app object, but we can also just access its elements at any time because its internal. I wish there was a good way to abstract it away a bit more

v3/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/newrelic/go-agent/v3
 go 1.20
 
 require (
-	google.golang.org/protobuf v1.5.3
+	github.com/golang/protobuf v1.5.3
```
iamemilio, line 6:
```suggestion
	google.golang.org/protobuf v1.5.3
```

iamemilio (dismissed):

nr-swilloughby:
Apparently my dev setup really wanted the go.mod to be sorted differently. It should be back to what's on the dev branch now.

Issue #950, opened 2024-08-19
integrations/nrzerolog imports error


Meroje:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your New Relic account or API keys! )

The new integrations/nrzerolog packages imports newrelic/go-agent in its testable example, instead of newrelic/go-agent/v3. This is causing errors with `go mod tidy`.

## Steps to Reproduce

[NOTE]: # ( Please be as specific as possible. )

Doesn't need to be in a new repo, just run mod tidy in the nrzerolog integration package.

```
$ go mod tidy
go: finding module for package github.com/newrelic/go-agent
go: github.m6web.fr/devops/go-fx-core/newrelic imports
	github.com/newrelic/go-agent/v3/integrations/nrzerolog tested by
	github.com/newrelic/go-agent/v3/integrations/nrzerolog.test imports
	github.com/newrelic/go-agent: module github.com/newrelic/go-agent@latest found (v3.34.0+incompatible), but does not contain package github.com/newrelic/go-agent
```

## Expected Behavior

[NOTE]: # ( Tell us what you expected to happen. )

## NR Diag results

[NOTE]: # ( Provide any other relevant log data. )

## Your Environment

[TIP]:  # ( Include as many relevant details about your environment as possible including the running version of New Relic software and any relevant configurations. )

## Reproduction case

[TIP]: # ( Link a sample application that demonstrates the issue. )

## Additional context

[TIP]:  # ( Add any other context about the problem here. )

nr-swilloughby:
Fixed in 3.35.0

PR #965, opened 2024-09-18
950 zerolog import fix

nr-swilloughby:
Fixes issue #950 
There was an error in the `import` statement in the example file for this integration.


iamemilio (approved):

Meroje:
why has this not been released, nrzerolog/1.0.0 is still faulty and should be retracted

Issue #906, opened 2024-04-29
Possible memory leak in Go agent


betabandido:
## Description

We have several Go applications that use the New Relic Go agent, and they work well. But there is a particular one that seems to be experiencing what looks like a memory leak. The setup for the agent is very similar in these applications. This might suggest that the problem is elsewhere, but we have ran Go's pprof tool and everything points to an issue with the NR Go agent. We're trying to create a reproducible example, but in the meantime we wanted to create this issue, and add to it as much information as possible, with the hope of setting the wheels turning to find a solution.

## Steps to Reproduce

We cannot reproduce the issue outside of this particular application (yet).

## Expected Behavior

Memory should not be increasing over time.

## NR Diag results

The following diagram shows the memory usage for the application running on our kubernetes cluster.

<img width="607" alt="image" src="https://github.com/newrelic/go-agent/assets/5481662/1b3d5379-bce2-497b-81f0-f9ae2f19262d">

As the image shows, memory keeps increasing until either we deploy a new release or the kubernetes cluster kills the app because of consuming too much memory.

## Your Environment

A Go application compiled using Go 1.21, running on a kubernetes cluster (EKS 1.27).

New Relic Go agent version 3.32.0.

## Reproduction case

We are working on trying to reproduce the issue.

## Additional context

Here you can see two outputs from pprof. The first one is from last Friday (Apr 26th), and the second one is from today (Apr 29th). As you can see the heap memory in use within `StoreLog` method has doubled (from ~40 MB to ~80 MB).

1. Apr 26th
![profile001](https://github.com/newrelic/go-agent/assets/5481662/082e8a1d-b67d-43e1-ad96-e6982f982a10)

2. Apr 29th
![profile002](https://github.com/newrelic/go-agent/assets/5481662/cddc645a-399e-453f-907b-dcc49ff71f1d)

iamemilio:
Hi @betabandido. Before we flag this as a memory leak, we need to inform you that as a feature of the go agent, transactions are stored on logs until that transaction ends. Transactions that run for a long time will accumulate all sorts of memory, but logs happen to be very large. We do this this way because of sampling in the agent. The weight of a sampled transaction is not calculated until it is finished, and so it puts all the data associated with it at risk until that point as well. So all logs associated with a transaction are stored on board until that transaction ends.

betabandido:
That makes sense. Yet, it seems we have none of the following:
- long-lasting transactions
- a continuously increasing number of transactions

[...]

PR #974, opened 2024-10-15
emergency memory cleanup of harvester

nr-swilloughby:
This is a feature addition I was holding back until I got confirmation that it was in fact the needed solution, but I think it may be best to go ahead and put it in as a default-off feature anyway to provide a means to mitigate a memory issue if one emerges with no other apparent cause or solution available and/or as a quick stop-gap solution for the customer until a better solution can be found.

This came out of the work on [Issue #906](https://github.com/newrelic/go-agent/issues/906), which was a reported memory leak apparently due to the agent holding onto log data longer than normal, but this only happens to this one application for this one customer under this one set of circumstances (a kubernetes environment) where memory constraints are a real issue.

Since one possibility for why it could be the case that log event data is retained longer than harvest cycles is a problem delivering them to the New Relic back-end collector (since the agent will wait for them to be delivered before dumping them), that might be a situation where a network issue or other external problem could indirectly cause the instrumented application to grow its memory too large to be viable.

And above all, we never want instrumentation of an application to unduly affect the operation of that application itself, so it stands to reason that if an application reaches that point where there seems like no other alternative, we should discard the accumulated event data in the harvester so the app can continue running.

This PR introduces an API call to allow the application to set a maximum heap size for the application. If it exceeds that value, all the harvester's data will be dropped and an emergency garbage collection and memory release will be requested. See the documentation for the function in the deltas for the PR for more details.

### DIFF
--- a/v3/newrelic/harvest.go
+++ b/v3/newrelic/harvest.go
@@ -4,6 +4,8 @@
 package newrelic
 
 import (
+	"runtime"
+	"runtime/debug"
 	"time"
 
 	"github.com/newrelic/go-agent/v3/internal"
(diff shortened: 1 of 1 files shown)


nr-swilloughby:
I think we should look at whether we want to allow more control over what memory is released here, since the only case we've found so far seems to be caused by memory issues outside the agent itself, and we're just providing a tool to help an application let go of resources to avoid a worse problem.

Issue #767, opened 2023-08-10
Heavily contended lock in TraceIDGenerator.generateID


ankon:
https://github.com/newrelic/go-agent/blob/046b4fc116adac2b4dec4d691b61a25b80488fc2/v3/internal/trace_id_generator.go#L52-L58

In a heavily used server a lot of segments end at the same time, and from what see in the profiles this lock is a problem.
1. Can it be avoided at all? For instance, using some magic "per-thread" generator (as much as that is possible, but possibly just having more than one rand instance might work?)
2. If it cannot be avoided, can it be made smaller? I see that it uses `defer` to unlock, which means it unlocks _after_ the conversion of the bits to the hex representation needed in the ID. Maybe it can be explicitly unlocked at least before that?

![image](https://github.com/newrelic/go-agent/assets/1210641/cfdf2025-753a-4ac7-9f06-e4fac7729a09)

ankon:
FWIW: I tried replacing the `defer tg.Unlock()` with an explicit unlock right after `tg.rnd.Read`, but benchmarks seem to imply that at least on my local machine that makes things slower.

What does help (see #977) is avoiding the allocation for the EncodeToString, and doing it in-place.

iamemilio:
I do think that an explicit unlock _may_ improve the async performance of the system as a whole, since the lock will not have to wait for the encoding process, even if it isn't necessarily faster for the performance of this function. According to your `pprof` output, it does look like a majority of the execution time is being spent on allocating and garbage collecting memory. We will take a look at your PR, I just enabled the tests.

ankon:
The fun one is: I benchmarked moving the unlock ... and it makes things slower.

```
goos: linux
goarch: amd64
pkg: github.com/newrelic/go-agent/v3/internal
cpu: AMD Ryzen 9 7940HS w/ Radeon 780M Graphics
BenchmarkTraceIDGenerator
BenchmarkTraceIDGenerator-16                    	25461540	        44.09 ns/op	      32 B/op	       2 allocs/op
BenchmarkTraceIDGeneratorParallel
BenchmarkTraceIDGeneratorParallel-16            	10721215	       107.5 ns/op	      32 B/op	       2 allocs/op
BenchmarkTraceIDGeneratorBaseline
BenchmarkTraceIDGeneratorBaseline-16            	24150187	        44.59 ns/op	      32 B/op	       2 allocs/op
BenchmarkTraceIDGeneratorBaselineParallel
[...]

PR #977, opened 2024-10-21
Improve performance of trace id generation

ankon:
## Links

Related-to: https://github.com/newrelic/go-agent/issues/767

## Details

Tried out various improvements for the linked issue, this PR improves the performance of the trace ID generation by doing the hex encoding in-place and avoiding an allocation. We see this function appearing in a high-throughput system, and hope that making it faster will also help us understand our own performance better.

benchstat for allocation avoidance:

```
goos: linux
goarch: amd64
pkg: github.com/newrelic/go-agent/v3/internal
cpu: AMD Ryzen 9 7940HS w/ Radeon 780M Graphics     
                            │ before.txt  │              after.txt              │
                            │   sec/op    │   sec/op     vs base                │
TraceIDGenerator-16           43.33n ± 3%   30.85n ± 1%  -28.81% (p=0.000 n=10)
TraceIDGeneratorParallel-16   71.23n ± 4%   60.40n ± 2%  -15.22% (p=0.000 n=10)
geomean                       55.56n        43.16n       -22.31%

                            │ before.txt │             after.txt              │
                            │    B/op    │    B/op     vs base                │
TraceIDGenerator-16           32.00 ± 0%   16.00 ± 0%  -50.00% (p=0.000 n=10)
TraceIDGeneratorParallel-16   32.00 ± 0%   16.00 ± 0%  -50.00% (p=0.000 n=10)
geomean                       32.00        16.00       -50.00%

                            │ before.txt │             after.txt              │
                            │ allocs/op  │ allocs/op   vs base                │
TraceIDGenerator-16           2.000 ± 0%   1.000 ± 0%  -50.00% (p=0.000 n=10)
TraceIDGeneratorParallel-16   2.000 ± 0%   1.000 ± 0%  -50.00% (p=0.000 n=10)
geomean                       2.000        1.000       -50.00%
```

### DIFF
--- a/v3/internal/trace_id_generator.go
+++ b/v3/internal/trace_id_generator.go
@@ -4,7 +4,6 @@
 package internal
 
 import (
-	"encoding/hex"
 	"math/rand"
 	"sync"
 )
@@ -39,6 +38,10 @@ const (
 	maxIDByteLen        = 16
 )
 
+const (
+	hextable = "0123456789abcdef"
+)
+
 // GenerateTraceID creates a new trace identifier, which is a 32 character hex string.
 func (tg *TraceIDGenerator) GenerateTraceID() string {
 	return tg.generateID(traceIDByteLen)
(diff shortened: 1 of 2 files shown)


iamemilio:
Out of curiosity, I wonder if just using the default seed and Rand source would be more efficient, since its already synchronized: https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/math/rand/rand.go;l=378

ankon:
I'm actually wondering whether we need the seeding at all. It is used for tests (fine), and otherwise just given the current timestamp. 

If that could be removed, one could start looking at other approaches here, for instance a pool of rand's, or a ringbuffer that gets filled independently and the generation just pulling out some bits ...

iamemilio (approved):

PR #978, opened 2024-10-24
hotfix security agent: only refresh state when connected to new relic

iamemilio:


v3/newrelic/secure_agent.go
```diff
@@ -43,7 +43,7 @@ func (app *Application) RegisterSecurityAgent(s securityAgent) {
 	if app != nil && app.app != nil && s != nil {
 		secureAgent = s
 		run, _ := app.app.getState()
-		if run != nil {
+		if run.Reply.IsConnectedToNewRelic() {
```
nr-swilloughby, line 46: Is there any case where `run` would be valid but `run.Reply` be `nil`? just wondering if we should make that `if run.Reply != nil && run.Reply.IsConnectedToNewRelic()`

nr-swilloughby (changes requested):

v3/newrelic/secure_agent.go
iamemilio: As the code is written now, we will always have either a real run.Reply, or a placeholder. We can check for to be defensive against future changes maybe?

mirackara (approved):

v3/newrelic/secure_agent.go
nr-swilloughby: probably wise

iamemilio: added

nr-swilloughby (approved):

PR #984, opened 2024-11-04
New config for nrsecurityagent

aayush-ap:
Added a new config to support the new feature release in CSEC Agent Release v1.5.0


v3/integrations/nrsecurityagent/nrsecurityagent.go
```diff
@@ -63,7 +69,7 @@ func InitSecurityAgent(app *newrelic.Application, opts ...ConfigOption) error {
 
 	appConfig, isValid := app.Config()
 	if !isValid {
-		return fmt.Errorf("Newrelic application value cannot be read; did you call newrelic.NewApplication?")
+		return fmt.Errorf("Newrelic  application value cannot be read; did you call newrelic.NewApplication?")
```
nr-swilloughby, line 72: Is that a typo or do you intend the extra space to be added to the error string?

nr-swilloughby (changes requested):
just one quick question, otherwise it looks good.

v3/integrations/nrsecurityagent/nrsecurityagent.go
aayush-ap: yes extra space is typo.

nr-swilloughby: ok we can fix it on release.

nr-swilloughby (approved):

Issue #986, opened 2024-11-19
nrslog.WithTransactionFromContext does not work after slog.With


moenth:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

When creating a `slog.Logger` using `nrslog.WithTransactionFromContext` transaction contexts stops being propagated after calling `With` on the resulting logger. This is because `TransactionFromContextHandler` does not implement its own `WithAttrs` and `WithGroup` functions which means they are delegated to the underlying `NRHandler`. Thus `TransactionFromContextHandler` is removed from the handler stack after calling either of these functions.

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your New Relic account or API keys! )

## Steps to Reproduce

```go
handler := nrslog.TextHandler(app, os.Stdout, &slog.HandlerOptions{})
logger := slog.New(nrslog.WithTransactionFromContext(handler))
appLogger := logger.With(slog.String("component", "app"))

tx := app.StartTransaction("test")
ctx := newrelic.NewContext(context.Background(), tx)

logger.InfoContext(ctx, "has transaction context")
appLogger.InfoContext(ctx, "does not have transaction context")
```

## Expected Behavior

Transaction context is propagated after using `slog.With`.

## NR Diag results

[NOTE]: # ( Provide any other relevant log data. )

## Your Environment

[TIP]:  # ( Include as many relevant details about your environment as possible including the running version of New Relic software and any relevant configurations. )

## Reproduction case

[TIP]: # ( Link a sample application that demonstrates the issue. )

## Additional context

[TIP]:  # ( Add any other context about the problem here. )

iamemilio:
Thank you for reporting this issue, we will make a ticket to address this ASAP.

iamemilio:
Solution in progress https://github.com/newrelic/go-agent/pull/997

iamemilio:
@moenth hopefully these changes will go live in the next week or two. We invite you to take a look if you would like.

iamemilio:
#997 has been merged and released

PR #987, opened 2024-11-20
Fix: nrslog integration with slog.Handler interface

iamemilio:
Fix for #986


iamemilio:
The fix to this is much less trivial than I expected. A mild rework is needed to correctly support attributes and groups.

PR #989, opened 2024-12-17
Slog with Groups and Attributes

mirackara:
Builds on-top of https://github.com/newrelic/go-agent/pull/987

- Support for transactions in context
```go
	txn := app.StartTransaction("test")
	logger.InfoContext(ctx, "log has tracing on new relic")
	txn.End()
``` 
- Corrected Grouping and Attributes System (Either Method Works)
```go
logger.WithGroup("program_info").With("pid", os.Getpid()).Info("log message with group and attribute")

logger.Info("I am a log group inside a log message",
	slog.Group("logGroup",
		slog.String("key1", "val"),
		slog.Int("key2", 1),
	),
)
```


mirackara:
Still Working on this - Updated the example main.go to show what is currently supported.

There are some bugs that still need squashing
- Issue with Attributes in the LogWithGroupAttributes example being Duplicated
- Issue with Attributes in the logWithLogAttrs example showing up correctly with their groups on nr1, but not in the log

PR #997, opened 2025-01-31
Slog 2.0

iamemilio:
A new approach to integrating New Relic with SLOG that is more lightweight, out of the way, and collects richer data. These changes have been constructed to be completely backwards compatible with v1 of nrslog. Changes include:
- A way to wrap slog.Handler objects with errors to allow users to handle invalid use cases
- A complete rework of log enrichment so that new relic linking metadata does not invalidate json, bson, or yaml scanners.  This new approach will instead inject the linking metadata as a key value pair. 
- Complete support for With(), WithGroup() and attributes for automatic instrumentation.
- A number of performance operations
- Robust testing at close to 90% coverage


v3/integrations/logcontext-v2/nrslog/attributes.go
```diff
@@ -0,0 +1,97 @@
+package nrslog
+
+import (
+	"log/slog"
+	"maps"
+	"strings"
+)
+
+type attributeCache struct {
+	preCompiledAttributes map[string]interface{}
```
nr-swilloughby, line 10: Recommend using `any` instead of `interface{}` these days as a style point, but I wouldn't hold up the review for that.

nr-swilloughby (approved):

Issue #995, opened 2025-01-30
No way to check if a transaction has ended


frknikiz:
[NOTE]: # ( Add a method to check if a transaction has ended )

## Summary

This feature request proposes adding a new method to the New Relic Go agent's `Transaction` object that allows developers to check if a transaction has been ended.

## Desired Behaviour

Currently, there is no way to directly determine if a transaction has been ended, especially when the `End()` call is made in a separate goroutine. This can lead to uncertainty and potential issues in tracking the status of transactions.

The desired behavior is to have a method, such as `IsEnded()`, that returns a boolean value indicating whether the `End()` method has been called on the transaction. This would allow developers to reliably check the transaction's status and handle it accordingly.

For example:

```go
txn:= newrelic.FromContext(r.Context())

//... some code...

go func() {
  //... some code...
  txn.End()
}()

//... some code...

if txn.IsEnded() {
  // Transaction has ended
} else {
  // Transaction is still active
}
```

## Possible Solution

Examining the current `Transaction` struct reveals a `finished` boolean field within the nested `txn` struct, which appears to track the transaction's end state. 

```go
type Transaction struct {
    Private any
    thread  *thread
}

type thread struct {
    *txn
    // ... other fields
}

type txn struct {
    // ... other fields
    finished           bool
    // ... other fields
}
```

Leveraging this existing field, the `IsEnded()` method could be implemented as follows:

```go
func (t *Transaction) IsEnded() bool {
  if t == nil || t.thread == nil || t.thread.txn == nil {
    return true
  }
  return t.thread.txn.finished
}
```

This method first handles potential nil cases to prevent panics and then directly accesses the `finished` field to provide the desired functionality.

## Additional context

This feature is important for applications that heavily rely on asynchronous processing and goroutines. In such scenarios, it can be difficult to keep track of the status of transactions, and the proposed method would provide a clear and reliable way to do so. This would help in debugging, ensuring proper transaction completion, and managing resources effectively.

mirackara:
Hey @frknikiz! We just released your contribution in v3.38.0. Closing this issue out now.

PR #998, opened 2025-02-01
feat(transaction-is-ended): add IsEnded() method (#995)

frknikiz:
## Links

- Issue: [#995](https://github.com/newrelic/go-agent/issues/995)

## Details

This pull request introduces a new `IsEnded()` method for the `Transaction` type. It addresses the issue raised in [#995](https://github.com/newrelic/go-agent/issues/995) where there was no straightforward way to determine if a transaction has ended.

**Changes include:**

- **New Method Implementation:**  
  The `IsEnded()` method has been added to check the internal transaction state. It ensures that a transaction is considered ended if it is nil, if its thread is nil, or if the underlying transaction's `finished` flag is set.

  ```go
  func (t *Transaction) IsEnded() bool {
      if t == nil || t.thread == nil || t.thread.txn == nil {
          return true
      }
      return t.thread.txn.finished
  }


- **Unit Tests:**

  Unit tests have been added to verify that:
    - A nil Transaction returns true for IsEnded().
    - A Transaction with a nil thread or nil txn returns true.
    - An active transaction returns false.
    - A finished transaction returns true.

These changes ensure robust transaction state checking across different scenarios.
Closes [#995](https://github.com/newrelic/go-agent/issues/995).

### DIFF
--- a/v3/newrelic/internal_txn.go
+++ b/v3/newrelic/internal_txn.go
@@ -54,6 +54,14 @@ type thread struct {
 	thread *tracingThread
 }
 
+func (thd *thread) IsEnded() bool {
+	txn := thd.txn
+	txn.Lock()
+	defer txn.Unlock()
+
+	return txn.finished
+}
+
 func (txn *txn) markStart(now time.Time) {
 	txn.Start = now
 	// The mainThread is considered active now.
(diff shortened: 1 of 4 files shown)


iamemilio:
Thanks for the PR, I like the idea. I would like to spend a little time next week to vet that this will be thread safe before moving forward with these changes, but I enabled the automated tests.

frknikiz:
The IsEnded method gives us information about whether the transaction has actually ended or not. For this reason, I think this error should not be returned. 

If the transaction is somehow nil,i think it's fine then the transaction is finished.

frknikiz:
Thanks for your support 🙏

PR #999, opened 2025-02-10
Go Chi Support

mirackara:
Go Chi Integration Support via Middleware


iamemilio (approved):
This looks good to me

PR #1001, opened 2025-02-12
prevent all transaction methods from nil transaction errors

iamemilio:
These changes add logic and tests that verify that all methods called on new relic transactions gracefully return when the transaction object is nil.

### DIFF
--- a/v3/newrelic/transaction.go
+++ b/v3/newrelic/transaction.go
@@ -24,11 +24,16 @@ type Transaction struct {
 	thread  *thread
 }
 
+// nilTransaction guards against nil errors when handling a transaction.
+func nilTransaction(txn *Transaction) bool {
+	return txn == nil || txn.thread == nil || txn.thread.txn == nil
+}
+
 // End finishes the Transaction.  After that, subsequent calls to End or
 // other Transaction methods have no effect.  All segments and
 // instrumentation must be completed before End is called.
 func (txn *Transaction) End() {
-	if txn == nil || txn.thread == nil {
+	if nilTransaction(txn) {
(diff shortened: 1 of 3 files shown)


mirackara (dismissed):

mirackara (approved):

v3/newrelic/transaction.go
```diff
@@ -324,7 +333,7 @@ func (txn *Transaction) startSegmentAt(at time.Time) SegmentStartTime {
 //	// ... code you want to time here ...
 //	segment.End()
 func (txn *Transaction) StartSegment(name string) *Segment {
-	if IsSecurityAgentPresent() && txn != nil && txn.thread != nil && txn.thread.thread != nil && txn.thread.thread.threadID > 0 {
+	if IsSecurityAgentPresent() && nilTransaction(txn) && txn.thread.thread.threadID > 0 {
 		// async segment start
```
nr-swilloughby, line 337: Shouldn't this be `!nilTransaction(txn)`?

nr-swilloughby (changes requested):

v3/newrelic/transaction.go
nr-swilloughby: Also, `nilTransaction(txn)` doesn't test for `txn.thread.thread != nil`

iamemilio: didnt know that was possible 😮 
--> will update

iamemilio: how does this look?

nr-swilloughby (approved):

PR #1003, opened 2025-02-25
Release 3.37.0 (version set and final changes)

nr-swilloughby:
### Enhanced
  - Implemented a new approach to integrating New Relic with SLOG that is more lightweight, out of the way, and collects richer data. These changes have been constructed to be completely backwards-compatible with v1 of nrslog. Changes include:
     - Wrapping `slog.Handler` objects with errors to allow users to handle invalid use cases
     - A complete rework of log enrichment so that New Relic linking metadata does not invalidate JSON, BSON, or YAML scanners. This new approach will instead inject the linking metadata as a key-value pair.
     - Complete support for `With()`, `WithGroup()`, and attributes for automatic instrumentation.
     - Performance operations.
     - Robust testing (close to 90% coverage).
     - **This updates logcontext-v2/nrslog to v1.4.0.**
  - Now custom application tags (labels) may be added to all forwarded log events.
     - Enabled if `ConfigAppLogForwardingLabelsEnabled(true)` or `NEW_RELIC_APPLICATION_LOGGING_FORWARDING_LABELS_ENABLED=TRUE`
     - May exclude labels named in `ConfigAppLogForwardingLabelsExclude("label1","label2",...)` or `NEW_RELIC_APPLICATION_LOGGING_FORWARDING_LABELS_EXCLUDE="label1,label2,..."`
     - Labels are defined via `ConfigLabels(...)` or `NEW_RELIC_LABELS`
  - Added memory allocation limit detection/response mechanism to facilitate calling custom functions to perform application-specific resource management functionality, report custom metrics or events, or take other appropriate actions, in response to rising heap memory size.

### Support statement
We use the latest version of the Go language. At minimum, you should be using no version of Go older than what is supported by the Go team themselves.
See the [Go agent EOL Policy](/docs/apm/agents/go-agent/get-started/go-agent-eol-policy) for details about supported versions of the Go agent and third-party components.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,25 @@
+## 3.37.0
+### Enhanced
+  - Implemented a new approach to integrating New Relic with SLOG that is more lightweight, out of the way, and collects richer data. These changes have been constructed to be completely backwards-compatible with v1 of nrslog. Changes include:
+     - Wrapping `slog.Handler` objects with errors to allow users to handle invalid use cases
+     - A complete rework of log enrichment so that New Relic linking metadata does not invalidate JSON, BSON, or YAML scanners. This new approach will instead inject the linking metadata as a key-value pair.
+     - Complete support for `With()`, `WithGroup()`, and attributes for automatic instrumentation.
+     - Performance operations.
+     - Robust testing (close to 90% coverage).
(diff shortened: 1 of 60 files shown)


iamemilio (approved):

iamemilio:
LGTM, good catch on the `nil` segment issue

PR #1006, opened 2025-03-18
Attribute value limit increase

iamemilio:
Dynamic Increases in Attribute Limits


nr-swilloughby (changes requested):
are we waiting for spec approval before merging this?

PR #1014, opened 2025-03-21
Update nrslog example

frankywahl:
nrslog in logs in context used a deprecated version of the nrslog API. We update the example to reflect better how to use it using the new API so that users can reference it and integrate the new API in their code.


iamemilio (approved):
Thanks for catching this!

nr-swilloughby (approved):

nr-swilloughby (approved):

PR #1021, opened 2025-04-07
support time objects for attribute values

### LINKED ISSUE #1013: nrslog: Logs in context does not work with durations
## Description

Structured logging in go can use the nrslog.Duration type to define an attribute. However, that does not get reported as a duration, but instead gets reported as the string `time.Duration`

## Steps to Reproduce

The following code sample is based from the example in https://github.com/newrelic/go-agent/blob/master/v3/integrations/logcontext-v2/nrslog/example/main.go and adds a `slog.Duration` attribute when logging. 

```go
package main

import (
	"context"
	"log/slog"
	"os"
	"time"

	"github.com/newrelic/go-agent/v3/integrations/logcontext-v2/nrslog"
)

func main() {
	app, err := newrelic.NewApplication(
		newrelic.ConfigAppName("slog example app"),
		newrelic.ConfigFromEnvironment(),
		newrelic.ConfigAppLogEnabled(true),
	)
	if err != nil {
		panic(err)
	}

	app.WaitForConnection(time.Second * 5)
	log := slog.New(nrslog.TextHandler(app, os.Stdout, &slog.HandlerOptions{}))

	log.Info("I am a log message")

	txn := app.StartTransaction("example transaction")

	log.InfoContext(ctx, "I am a log inside a transaction with custom attributes!",
		slog.String("foo", "bar"),
		slog.Int("answer", 42),
		slog.Duration("duration", 3*time.Second),
		slog.Any("some_map", map[string]interface{}{"a": 1.0, "b": 2}),
	)

	// pretend to do some work
	time.Sleep(500 * time.Millisecond)
	log.Warn("Uh oh, something important happened!")
	txn.End()

	log.Info("All Done!")

	app.Shutdown(time.Second * 10)
}
```

## Expected Behavior

An actual duration (as integer, or float based on the unit) would be populated instead of the string. 

## Your Environment

Go: 1.24
Go-Agent: v3.38.0

## Reproduction case

Image

## Additional context

Note: `time.Time` has similar issues in that it comes as a string timestamp which NR would then have to parse.

iamemilio:
fix #1013

### DIFF
--- a/v3/newrelic/attributes_test.go
+++ b/v3/newrelic/attributes_test.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/newrelic/go-agent/v3/internal/crossagent"
 )
@@ -170,6 +171,7 @@ func TestWriteAttributeValueJSON(t *testing.T) {
 	writeAttributeValueJSON(&w, "a", int(-5))
 	writeAttributeValueJSON(&w, "a", float32(1.5))
 	writeAttributeValueJSON(&w, "a", float64(4.56))
+	writeAttributeValueJSON(&w, "duration", time.Duration(7))
 	buf.WriteByte('}')
 
 	expect := compactJSONString(`{
@@ -188,7 +190,8 @@ func TestWriteAttributeValueJSON(t *testing.T) {
 		"a":-4,
 		"a":-5,
 		"a":1.5,
-		"a":4.56
+		"a":4.56,
+		"duration":"7ns"
 		}`)
 	js := buf.String()
 	if js != expect {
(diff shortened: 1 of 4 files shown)


v3/newrelic/attributes_test.go
```diff
@@ -188,7 +190,8 @@ func TestWriteAttributeValueJSON(t *testing.T) {
 		"a":-4,
 		"a":-5,
 		"a":1.5,
-		"a":4.56
+		"a":4.56,
+		"duration":"7ns"
 		}`)
```
nr-swilloughby, line 195: Curious, why not have test cases for the other new time units too?

nr-swilloughby (changes requested):

nr-swilloughby (approved):

PR #1036, opened 2025-05-29
feat: Improve developer experience

bduranleau-nr:
++ Adds a docker-compose.yml file that will orchestrate the container environment
++ Adds a Makefile with several development and testing targets
++ Adds a new test-pull-request.yml GHA workflow
++ Fixes `nrpgx5` integration tests
++ Adds a developer guide section to the README

-- removes ci.yaml GHA workflow


Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-420447

.github/workflows/test-pull-request.yml
```diff
@@ -0,0 +1,178 @@
...
+          make test-services-start
+      - name: Run Integration tests
+        working-directory: ./go-agent
+        run: |
+          docker exec -e TEST=${{ matrix.integration-test }} nr-go make integration-test
+      - name: Stop services
+        working-directory: ./go-agent
+        run: |
+          make test-services-stop
+
```
mirackara, line 170: I think the integration tests here are missing the codecov uploading step like in the core-tests above

bduranleau-nr: Added codecov to the amd64 integration tests here: 86df483fd6c5a2d5913029261161212546edbfdd
Do we also want to run codecov against the arm64 integration tests?

mirackara: I don't think we need to. We can save on time if we only have amd64 codecov reporting since the coverage reports would be identical

README.md
```diff
@@ -168,9 +168,214 @@ If you are already using another open source solution to gather telemetry data,
...
+
+[Docker Compose](https://docs.docker.com/compose/) builds on the functionality provided by Docker to control the orchestration of multi-container environments.
+
+The development workflow described in this section relies heavily Docker. The containers that are generated will create a [volume](https://docs.docker.com/engine/storage/volumes/) mounted to the current directory on the local host, allowing the modification of code and tests on the local host to be propagated automatically to the running container without having to re-build or copy manually.
+
+### Targets
+
+| Target | Description |
+|--------|-------------|
+| `dev-image` | Build the dev image(s) specified under the `docker compose` `dev` profile. |
```
mvicknr: this looks like it was renamed to `devenv-image`

README.md
```diff
@@ -168,9 +168,214 @@ If you are already using another open source solution to gather telemetry data,
...
+| `dev-shell` | Run `docker compose` with the `dev` profile. Open a shell in the running container. |
+| `dev-stop`  | Terminate the running containers under the `docker compose` `dev` profile. |
+| `test-services-start` | Spin up agent and other containers. |
+| `test-services-stop`  | Terminate agent and other containers. |
+| `core-test`  | Run a specified core test. |
+| `core-suite` | Run all core tests. |
+| `integration-test`  | Run a specified integration test. |
+| `integration-suite` | Run all integration tests. |
```
mvicknr, line 199: I ran `make test-services-start` and then `integration-suite` and got this failure. Is that expected? It did not fail when I ran it is the dev-shell. I also tried `make test-services-start PROFILE=nrpgx5` and then `integration-suite` and got this same error.

```
# TEST=nrpgx5, COVERAGE=
--- FAIL: TestTracer_Trace_CRUD (0.01s)
    --- FAIL: TestTracer_Trace_CRUD/query_should_send_the_metric_after_the_row_close (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x1015b9fbc]
```

README.md
```diff
@@ -168,9 +168,214 @@ If you are already using another open source solution to gather telemetry data,
...
+make dev-shell
+```
+
+Terminate the container(s) started by `dev-shell`:
+```
+make dev-stop
+```
+
+Spin up services for testing. Does not create a shell into the container.
+May specify an optional `PROFILE` argument to start another container not covered by the default `test` profile:
```
mvicknr, line 216: It may be a good idea to add the note from line 318 here as well

README.md
bduranleau-nr: Are you just running `integration-suite`, or are you following the instructions laid out in [this section](https://github.com/newrelic/go-agent/pull/1036/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R339) for running tests against the test-services target using `docker exec`? I can't reproduce this error when following those instructions.

README.md
bduranleau-nr: This is intended to be a more generalized statement to account for other profiles that may be added in the future (like one for mongodb). 318 is more specific to the `nrpgx5` example test being run there.

README.md
bduranleau-nr: fixed in a734bc7146304ec9c87dd2ea007b99a5c45c3dd5

README.md
mvicknr: I was able to get it to pass by running
`make test-services-start`
and then 
`docker exec -e TEST=nrpgx5 nr-go make integration-test`
I had a bit of trouble understanding what order to do things in and what was required for each make command.

README.md
mvicknr: Ok, I was thinking that if you want to run all of the integration tests at once, they will fail if you don't add that profile but perhaps that isn't the case?

README.md
bduranleau-nr: in order to run the full suite rather than just an individual test, you can leave out the env var arg and just use `docker exec nr-go make integration-suite`. Does this work?

> I had a bit of trouble understanding what order to do things in and what was required for each make command.

I won't deny there's a bit of a learning curve. I think `test-services` will mostly be used for the GHA workflow, or for someone who wants to test multiple Go versions locally in parallel. `dev-shell` is probably going to be the more used target for local development, and is a bit easier to navigate the commands for.

README.md
bduranleau-nr: I see what you mean- I'll add a callout that `integration-suite` will require specifying all PROFILEs at build time.

bduranleau-nr: This also makes me think that we'll need a target for the `integration-suite` tests that spins up every dependent container when we do add other integrations like mongodb, since right now the assumption is that only one extra profile will be needed. Perhaps a `suite` `PROFILE`?

bduranleau-nr: For context: the reason I've broken it out to not always just spin up the accessory containers is for GHA efficiency- to avoid pulling and building the postgres image for every integration test when we only need it for `nrpgx5`.

README.md
mvicknr: Yes, `docker exec nr-go make integration-suite` works for me. I agree, I'll probably use `dev-shell` and was just testing everything out to make sure I understood it and could provide a meaningful PR review.

README.md
bduranleau-nr: My core assumption was that 
1. `test-services` wouldn't be used very often locally
2. Someone working on the agent would probably only have to spin up at most one integration container at a time for testing. Using mongodb as an example, someone wouldn't need the postgres container if they were only working on the `nrmongo` integration, assuming they wanted to use the `test-services` targets. The `dev-shell` target will always spin up every container, just for ease of use.

bduranleau-nr: I've added a callout for this scenario in this commit 8288f5cc3e169725dd3a7f5e1a57e0683853f8f0

mvicknr (approved):
LGTM

PR #1041, opened 2025-06-09
feat: Add logging custom attributes

yahmed153:
## Links

https://new-relic.atlassian.net/browse/NR-354877

## Details
Added Custom Attributes via environment variable `NEW_RELIC_APPLICATION_LOGGING_FORWARDING_CUSTOM_ATTRIBUTES` or programmatically. Format of key/value is `key1:val1;key2:val2;....`

### DIFF
--- a/v3/newrelic/config.go
+++ b/v3/newrelic/config.go
@@ -640,7 +640,7 @@ func defaultConfig() Config {
 
 	c.Enabled = true
 	c.Labels = make(map[string]string)
-	c.CustomAttributes = nil
+	c.CustomAttributes = make(map[string]string)
 	c.CustomInsightsEvents.Enabled = true
 	c.CustomInsightsEvents.MaxSamplesStored = internal.MaxCustomEvents
 	c.TransactionEvents.Enabled = true
@@ -836,6 +836,12 @@ func copyConfigReferenceFields(cfg Config) Config {
 		cp.Labels = make(map[string]string, len(cfg.Labels))
 		for key, val := range cfg.Labels {
 			cp.Labels[key] = val
+		}
+	}
+	if nil != cfg.CustomAttributes {
+		cp.CustomAttributes = make(map[string]string, len(cfg.CustomAttributes))
+		for key, val := range cfg.CustomAttributes {
+			cp.CustomAttributes[key] = val
 		}
 	}
 	if cfg.ErrorCollector.IgnoreStatusCodes != nil {
--- a/v3/newrelic/config.go
+++ b/v3/newrelic/config.go
@@ -578,6 +578,8 @@ type ApplicationLogging struct {
 			// List of label types to exclude from forwarded logs.
 			Exclude []string
 		}
+		// Key/Value pairs to add as custom attributes to all log events forwarded to New Relic.
+		CustomAttributes map[string]string
 	}
 	Metrics struct {
 		// Toggles whether the agent gathers the the user facing Logging/lines and Logging/lines/{SEVERITY}
--- a/v3/newrelic/config.go
+++ b/v3/newrelic/config.go
@@ -48,11 +48,6 @@ type Config struct {
 	// https://docs.newrelic.com/docs/using-new-relic/user-interface-functions/organize-your-data/labels-categories-organize-apps-monitors
 	Labels map[string]string
 
-	// CustomAttributes A hash with key/value pairs to add as custom attributes to all log events forwarded to New Relic.
-	//
-	// https://docs.newrelic.com/docs/TODO
-	CustomAttributes map[string]string
-
 	// HighSecurity guarantees that certain agent settings can not be made
 	// more permissive.  This setting must match the corresponding account
 	// setting in the New Relic UI.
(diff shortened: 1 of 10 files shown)


v3/newrelic/config_options.go
```diff
@@ -389,46 +389,54 @@ func ConfigLabels(labels map[string]string) ConfigOption {
...
+//		NEW_RELIC_APPLICATION_LOGGING_FORWARDING_LABELS_ENABLED     sets ApplicationLogging.LogForwarding.Labels.Enabled to enable sending application labels with forwarded logs.
+//		NEW_RELIC_APPLICATION_LOGGING_FORWARDING_LABELS_EXCLUDE     sets ApplicationLogging.LogForwarding.Labels.Exclude to filter out a set of unwanted label types from the ones reported with logs.
+//	 	NEW_RELIC_APPLICATION_LOGGING_METRICS_ENABLED		  		sets ApplicationLogging.Metrics.Enabled. Set to false to disable the collection of application log metrics.
+//	 	NEW_RELIC_APPLICATION_LOGGING_LOCAL_DECORATING_ENABLED      sets ApplicationLogging.LocalDecoration.Enabled. Set to true to enable local log decoration.
+//		NEW_RELIC_APPLICATION_LOGGING_FORWARDING_MAX_SAMPLES_STORED	sets ApplicationLogging.LogForwarding.Limit. Set to 0 to prevent captured logs from being forwarded.
+//		NEW_RELIC_APPLICATION_LOGGING_FORWARDING_CUSTOM_ATTRIBUTES	sets ApplicationLogging.Forwarding.CustomAttributes A hash with key/value pairs to add as custom attributes to all log events forwarded to New Relic.
```
yahmed153: added via env var

yahmed153:
TODO: Add link to docs like [Labels](https://docs.newrelic.com/docs/using-new-relic/user-interface-functions/organize-your-data/labels-categories-organize-apps-monitors)

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-423812

yahmed153:
@nr-swilloughby would we need to add `CustomAttributes` to this [struct](https://github.com/newrelic/go-agent/blob/52665e5a9484af02a5a5d145541278142956ccee/v3/newrelic/config.go#L969)

nr-swilloughby:
> @nr-swilloughby would we need to add `CustomAttributes` to this [struct](https://github.com/newrelic/go-agent/blob/52665e5a9484af02a5a5d145541278142956ccee/v3/newrelic/config.go#L969)

I don't think we necessarily do, but that's a good question. We don't normally add things there for every feature we add. The items there are specifically the values that show up at the top level of the configuration values for the agent's connect payload, and are extracted from the `Config` struct. However, as you can see the entire `Config` struct is also included as the `settings` value of that data structure itself anyway, so your values are already in there automatically, just deeper down inside the struct. So unless there's a specific reason to bring a value up to the top level, we don't need to do anything and they'll just show up in the general settings value.

v3/newrelic/config.go
```diff
@@ -827,6 +838,12 @@ func copyConfigReferenceFields(cfg Config) Config {
 			cp.Labels[key] = val
 		}
 	}
+	if nil != cfg.CustomAttributes {
```
nr-swilloughby: Please don't use "yoda" conditionals. "if cfg.CustomAttributes != nil" is the preferred way to state the expression.

v3/newrelic/config.go
```diff
@@ -48,6 +48,11 @@ type Config struct {
 	// https://docs.newrelic.com/docs/using-new-relic/user-interface-functions/organize-your-data/labels-categories-organize-apps-monitors
 	Labels map[string]string
 
+	// CustomAttributes A hash with key/value pairs to add as custom attributes to all log events forwarded to New Relic.
+	//
+	// https://docs.newrelic.com/docs/TODO
+	CustomAttributes map[string]string
```
nr-swilloughby: I would recommend moving this so all of the options for the feature are located together (the enable boolean and this map), either as a pair of values in `CustomInsightsEvents` like
```
CustomInsightsEvents struct {
   CustomAttributesEnabled bool
   CustomAttributesValues map[string]string
}
```
or as a sub-struct like
```
CustomInsightsEvents struct {
   CustomAttributes struct {
      Enabled bool {
      Values map[string]string
   }
}
```
(or whatever is most appropriate for naming instead of `Values` there), so they're not separated from each other in the `Config` struct.

v3/newrelic/log_events.go
```diff
@@ -181,6 +181,14 @@ func (events *logEvents) CollectorJSON(agentRunID string) ([]byte, error) {
 			}
 		}
 	}
+	if events.config.customAttributes != nil {
+		for k, v := range events.config.customAttributes {
+			buf.WriteByte(',')
+			jsonx.AppendString(buf, "tags."+k)
```
nr-swilloughby: Is this what the spec for custom attributes is asking for? I think the `tags.` prefix was specifically for labels. custom log events would be something else, otherwise the feature would be identical to labels, wouldn't it?

v3/newrelic/config.go
nr-swilloughby: I know we have these in our legacy code, but we're trying to correct them as we find them and not put new ones in when writing new code :)

nr-swilloughby (changes requested):

v3/newrelic/config.go
yahmed153: lol will do, agree

v3/newrelic/log_events.go
yahmed153: There is no mention of a prefix in the [spec](https://source.datanerd.us/agents/agent-specs/pull/709/files) to custom attributes. Will remove the prefix `tags`

nr-swilloughby (approved):

PR #1043, opened 2025-06-09
chore(v3): go fmt codebase

bduranleau-nr:


v3/newrelic/tracing_test.go
```diff
@@ -221,11 +222,12 @@ func TestLostChildren(t *testing.T) {
 	})
 }
 
-//                                          |-t3-|    |-t4-|
-//                           |-t2-|    |-never-finished----------
-//            |-t1-|    |--never-finished------------------------
-//  |-------root-------------------------------------------------
-//  0    1    2    3    4    5    6    7    8    9    10   11   12
+//	                              |-t3-|    |-t4-|
```
mvicknr: is the meaning is this changed? for example t1 used to span between 2 and 3 but now it spans between 1 and 2

bduranleau-nr: Good catch - the go fmt auto formatter doesn't like the blank space in the comment, and inserts tab characters at the beginning of the line. We'll have to think of a way around this since there isn't an option to ignore blocks of code for go fmt that I'm aware of.

bduranleau-nr: 4c647e7ab9c3de74570df0c0ea1e141300624d59 should restore the correct alignment

mvicknr: latest change LGTM

mvicknr (dismissed):

mvicknr (approved):

PR #1049, opened 2025-06-20
feat: mongodb v2 integration

mvicknr:
## Links

https://new-relic.atlassian.net/browse/NR-404946

## Details

Adding v2 Mongo integration https://pkg.go.dev/go.mongodb.org/mongo-driver/v2
Updating v1 Mongo integration to latest version
Cleaning up `integration-tests.mk`


integration-tests.mk
mvicknr, line 1: I reordered to alphabetic order, added missing dirs, removed dirs with no tests, and added `nrmongo-v2`

v3/integrations/nrmongo-v2/go.mod
```diff
@@ -0,0 +1,29 @@
...
+// https://github.com/mongodb/mongo-go-driver#requirements
+go 1.22
+
+require (
+	github.com/newrelic/go-agent/v3 v3.39.0
+	// mongo-driver does not support modules as of Nov 2019.
+	go.mongodb.org/mongo-driver/v2 v2.2.2
+)
+
+require (
```
mvicknr: I tried removing this section since other go.mod files don't have it but I can't commit without it

mvicknr:
I am looking at using a real Mongo db to test (like postgres does) but if that is done, it will be a separate PR.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-429069

integration-tests.mk
bduranleau-nr: I would actually suggest against removing directories with no tests- this will have an impact to our codecoverage statistics and give an inaccurate representation of our coverage metrics. This list is what is used to generate the coverage profiles for each of the packages/directories

v3/integrations/nrmongo/nrmongo_test.go
```diff
@@ -28,10 +28,10 @@ var (
 		ConnectionID: connID,
 	}
 	finishedEvent = event.CommandFinishedEvent{
-		DurationNanos: 5,
-		CommandName:   "name",
-		RequestID:     reqID,
-		ConnectionID:  connID,
+		Duration:     5,
```
mvicknr, line 31: Fixing deprecated field

integration-tests.mk
mvicknr: That makes sense. There are some nuances like these examples....

Here is a case where the example folder has a test, but most do not:
https://github.com/newrelic/go-agent/blob/mvick/mongodb-upgrades/v3/integrations/nrfiber/example/main_test.go

Here is an example where the test is in a separate folder:
https://github.com/newrelic/go-agent/blob/mvick/mongodb-upgrades/v3/integrations/nrstan/test/nrstan_test.go

Perhaps we don't include any child folders in coverage, just the root folder (except in the case of logcontext folders and the nrstan special case)? cc @mirackara @nr-swilloughby

bduranleau-nr: That's a strange one- I don't think that files under `example` folders should have tests/be included in test coverage metrics at all really. Currently nothing except for explicitly defined directories in the `-tests.mk` files should not be tested/included in code coverage metrics, since the test command only specifies operating against the `./` directory: https://github.com/newrelic/go-agent/blob/develop/Makefile#L65

nr-swilloughby: interesting question. On the one hand, it's good for examples to show tests so we're demonstrating good examples and also how to have testable example apps but also it seems like we should be running those tests to as a test to make sure our examples keep working, shouldn't we?

mvicknr: The team had a discussion and we will include all of the root directories under `integrations` but not the child directories like `example` and `test`

v3/integrations/nrmongo-v2/example/main.go
```diff
@@ -0,0 +1,53 @@
...
+	"go.mongodb.org/mongo-driver/v2/mongo/options"
+)
+
+func main() {
+	app, err := newrelic.NewApplication(
+		newrelic.ConfigAppName("Basic Mongo App"),
+		newrelic.ConfigLicense(os.Getenv("NEW_RELIC_LICENSE_KEY")),
+		newrelic.ConfigDebugLogger(os.Stdout),
+	)
+	if nil != err {
```
bduranleau-nr: Avoid yoda conditionals
```suggestion
	if err != nil {
```

v3/integrations/nrmongo-v2/example/main.go
```diff
@@ -0,0 +1,53 @@
...
+		panic(err)
+	}
+	app.WaitForConnection(10 * time.Second)
+
+	// If you have another CommandMonitor, you can pass it to NewCommandMonitor and it will get called along
+	// with the NR monitor
+	nrMon := nrmongo.NewCommandMonitor(nil)
```
bduranleau-nr: Suggestion: make clear what `nrMon` is here by changing the name slightly. It could be construed as `nrMongo` given the context of this code. Since it's a monitor, `nrMonitor` or `nrCmdMonitor` would probably be best to avoid confusion.

v3/integrations/nrmongo-v2/nrmongo.go
```diff
@@ -0,0 +1,196 @@
...
+//	}
+//
+//	// Add transaction to the context.  This step is required.
+//	ctx = newrelic.NewContext(ctx, txn)
+//
+//	collection := client.Database("testing").Collection("numbers")
+//	resp, err := collection.InsertOne(ctx, bson.M{"name": "pi", "value": 3.14159})
+//	if err != nil {
+//		log.Fatal(err)
+//	}
```
bduranleau-nr: This block seems to be slightly at odds with our `main.go` example. I would suggest removing this example code entirely from this comment block to protect against the code and documentation getting out of sync.

v3/integrations/nrmongo-v2/nrmongo.go
```diff
@@ -0,0 +1,196 @@
+// Copyright 2020 New Relic Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package nrmongo instruments https://github.com/mongodb/mongo-go-driver v2
```
bduranleau-nr, line 4: is the `v2` at the end of this comment line supposed to be part of the URL?

v3/integrations/nrmongo-v2/nrmongo.go
```diff
@@ -0,0 +1,196 @@
...
+//		Succeeded: origSucceeded,
+//		Failed:    origFailed,
+//	}
+//	nrMon := nrmongo.NewCommandMonitor(origMon)
+//	client, err := mongo.Connect(options.Client().SetMonitor(nrMon))
+//
+// Then add the current transaction to the context used in any MongoDB call:
+//
+//	ctx = newrelic.NewContext(context.Background(), txn)
+//	resp, err := collection.InsertOne(ctx, bson.M{"name": "pi", "value": 3.14159})
```
bduranleau-nr: This block seems to be slightly out of sync with our `main.go` example

v3/integrations/nrmongo-v2/nrmongo.go
mvicknr: no, the same repo contains all versions

v3/integrations/nrmongo-v2/nrmongo_test.go
```diff
@@ -0,0 +1,261 @@
...
+	txn = app.StartTransaction("txnName")
+	ctx = newrelic.NewContext(context.Background(), txn)
+	nrMonitor.started(ctx, ste)
+	if len(nrMonitor.segmentMap) != 1 {
+		t.Errorf("Wrong number of segments, expected 1 but got %d", len(nrMonitor.segmentMap))
+	}
+	nrMonitor.failed(ctx, fe)
+	if !failed {
+		t.Error("Original monitor not succeeded")
+	}
```
bduranleau-nr: So if `failed` is `false`, this test fails? Is the error message correct? It looks like it's copied from the `succeeded` test above. What is the context for the `nrMonitor.failed` function?

v3/integrations/nrmongo-v2/nrmongo.go
```diff
@@ -0,0 +1,196 @@
...
+		Collection:   collName(e),
+		Operation:    e.CommandName,
+		Host:         host,
+		PortPathOrID: port,
+		DatabaseName: e.DatabaseName,
+	}
+	if newrelic.IsSecurityAgentPresent() {
+		sgmt.SetSecureAgentEvent(secureAgentEvent)
+	}
+	m.addSgmt(e, &sgmt)
```
bduranleau-nr, line 112: It looks like this functionality is not captured by any of the current tests- instead we just check to see if the original monitor was called. It would probably be good to check that this functionality is also working as expected.

bduranleau-nr: ah, I can see this is somewhat tested [here](https://github.com/newrelic/go-agent/pull/1049/files#diff-3a83ee6e9b8fd28217a0dcdb70d31a5c90c4fd310509d39df0755ab5dcfae7dcR186) by checking the `segmentMap` length.

v3/integrations/nrmongo-v2/nrmongo_test.go
```diff
@@ -0,0 +1,261 @@
...
+		origCommMon: origMonitor,
+	}
+	app := createTestApp()
+	txn := app.StartTransaction("txnName")
+	ctx := newrelic.NewContext(context.Background(), txn)
+	nrMonitor.started(ctx, ste)
+	if !started {
+		t.Error("Original monitor not started")
+	}
+	if len(nrMonitor.segmentMap) != 1 {
```
bduranleau-nr, line 123: I would suggest checking not just the length but the content of the `segmentMap` here, to make sure it isn't satisfying the length with unexpected/junk data.

v3/integrations/nrmongo-v2/nrmongo_test.go
```diff
@@ -0,0 +1,261 @@
...
+			},
+			UserAttributes:  map[string]interface{}{},
+			AgentAttributes: map[string]interface{}{},
+		},
+	})
+
+	txn = app.StartTransaction("txnName")
+	ctx = newrelic.NewContext(context.Background(), txn)
+	nrMonitor.started(ctx, ste)
+	if len(nrMonitor.segmentMap) != 1 {
```
bduranleau-nr, line 192: Same suggestion as https://github.com/newrelic/go-agent/pull/1049/files#r2162631711

v3/integrations/nrmongo-v2/nrmongo.go
mvicknr: What specifically isn't matching? This 2nd example uses the original monitor and our main is more like the first monitor on line 11. The other difference I see is passing the URI but I think that is just for a localhost example.

v3/integrations/nrmongo-v2/example/main.go
mvicknr: done in old and new version of mongo integrations

v3/integrations/nrmongo-v2/nrmongo.go
mvicknr: removed this extra example in both old and new versions of mongo integration

v3/integrations/nrmongo-v2/nrmongo_test.go
mvicknr: updated for clarity

v3/integrations/nrmongo-v2/nrmongo_test.go
mvicknr: done in both v1 and v2 versions

v3/integrations/nrmongo-v2/nrmongo_test.go
mvicknr: done in both v1 and v2 versions

v3/integrations/nrmongo-v2/nrmongo.go
bduranleau-nr: Should we include the other elements where we explicitly start & end the txn?

mvicknr: I don't think so since I don't see that included in the usage example of any of the other integrations

v3/integrations/nrmongo-v2/nrmongo_test.go
```diff
@@ -252,6 +264,39 @@ func TestGetJsonQuery(t *testing.T) {
 	}
 }
 
+func confirmSegValues(t *testing.T, seg *newrelic.DatastoreSegment) {
```
bduranleau-nr, line 267: Nice!

bduranleau-nr (approved):

nr-swilloughby (approved):
Looks great to me

PR #1052, opened 2025-06-28
fix user-controlled input internal_response_writer must be sanitized before being written to the HTTP response

odaysec:
https://github.com/newrelic/go-agent/blob/ac0e154a1445204debd8e87e182c0d698d2f996b/v3/newrelic/internal_response_writer.go#L29-L29
https://github.com/newrelic/go-agent/blob/ac0e154a1445204debd8e87e182c0d698d2f996b/v3/newrelic/internal_response_writer.go#L33-L33

directly writing user input (HTTP request parameter) to an HTTP response without properly sanitizing the input first, allows for a cross-site scripting vulnerability. this kind of vulnerability is also called reflected cross-site scripting, to distinguish it from other types of cross-site scripting.

fix `feat:internal_response_writer` HTTP response without properly sanitizing leading reflected cross site scripting, the user-controlled input (`b`) must be sanitized before being written to the HTTP response. The best approach is to use the `html.EscapeString` function from the `html` package to escape special characters in the input, ensuring that malicious scripts cannot be executed. This change should be applied in the `replacementResponseWriter.Write` method.

### Recommendation References
To guard against cross-site scripting, consider using contextual output encoding/escaping before writing user input to the response, or one of the other solutions that are mentioned in the references.
- [Types of Cross-Site Scripting](https://www.owasp.org/index.php/Types_of_Cross-Site_Scripting)
- [01aee9a394bf559ae350f3e34ede6712](https://gist.github.com/odaysec/01aee9a394bf559ae350f3e34ede6712)

### DIFF
--- a/v3/newrelic/internal_response_writer.go
+++ b/v3/newrelic/internal_response_writer.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"html"
 )
 
 type replacementResponseWriter struct {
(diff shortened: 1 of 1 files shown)


nr-swilloughby:
This code is correct as written. The HTTP data is being routed directly to our own web security framework specifically to detect this and other sorts of vulnerabilities as part of active security vulnerability scanning of instrumented apps.

PR #1073, opened 2025-07-28
feat: add remote parent sampling

mirackara:


Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-446927

v3/newrelic/config.go
```diff
@@ -279,6 +279,11 @@ type Config struct {
 		// ReservoirLimit sets the desired maximum span event reservoir limit
 		// for collecting span event data. The collector MAY override this value.
 		ReservoirLimit int
+
```
mirackara, line 290: Add Comment describing sampler config options

v3/newrelic/config_test.go
```diff
@@ -185,7 +185,7 @@ func TestCopyConfigReferenceFieldsPresent(t *testing.T) {
 					"Threshold":10000000
```
bduranleau-nr, line 185: I would suggest adding tests for setting the other 2 config options- one for `"always_on`" and one for `"always_off"`

mirackara: This is only for testing the existing default config output. I did just add a couple more tests to test the different config options however

v3/newrelic/internal_distributed_trace_test.go
```diff
@@ -1476,6 +1511,191 @@ func TestW3CTraceHeaders(t *testing.T) {
 
 }
 
+func TestW3CTraceHeadersSamplingDefault(t *testing.T) {
```
bduranleau-nr: It's a little difficult to parse the exact nature of these tests at a glance given the confusing nature of the spec- can you add a brief comment describing the exact test case being exercised by each test function?

PR #1076, opened 2025-08-01
test: Add secure agent tests and fix issues found with tests

mvicknr:
## Links
https://new-relic.atlassian.net/browse/NR-320174
https://new-relic.atlassian.net/browse/NR-446838

## Details

Found a logic issue in `Write` function - truncated cases were missing 1 character

### DIFF
--- a/v3/newrelic/secure_agent.go
+++ b/v3/newrelic/secure_agent.go
@@ -61,11 +61,9 @@ func getLinkedMetaData(app *app) map[string]string {
 	if app != nil && app.run != nil {
 		runningAppData["hostname"] = app.run.Config.hostname
 		runningAppData["entityName"] = app.run.firstAppName
-		if app.run != nil {
-			runningAppData["entityGUID"] = app.run.Reply.EntityGUID
-			runningAppData["agentRunId"] = app.run.Reply.RunID.String()
-			runningAppData["accountId"] = app.run.Reply.AccountID
-		}
+		runningAppData["entityGUID"] = app.run.Reply.EntityGUID
+		runningAppData["agentRunId"] = app.run.Reply.RunID.String()
+		runningAppData["accountId"] = app.run.Reply.AccountID
 	}
 	return runningAppData
 }
--- a/v3/newrelic/secure_agent_test.go
+++ b/v3/newrelic/secure_agent_test.go
@@ -97,6 +97,13 @@ func TestApplicationRegisterSecurityAgentNilCases(t *testing.T) {
 	originalAgent := secureAgent
 	defer func() { secureAgent = originalAgent }()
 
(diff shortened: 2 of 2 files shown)


Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-446915

v3/newrelic/secure_agent.go
```diff
@@ -113,16 +113,22 @@ type BodyBuffer struct {
 }
 
 func (b *BodyBuffer) Write(p []byte) (int, error) {
```
mvicknr, line 113: I flipped the logic here to exit right away if there is no room to write and to make it easier to read. If you'd rather have it closer to the original code, this would be the change:

```// to keep close to current logic
func (b *BodyBuffer) Write(p []byte) (int, error) {
	l := len(b.buf)
	if len(p) <= secureAgent.RequestBodyReadLimit()-l {
		b.buf = append(b.buf, p...)
		return len(p), nil
	} else if secureAgent.RequestBodyReadLimit()-l >= 1 {
		end := secureAgent.RequestBodyReadLimit() - l
		b.buf = append(b.buf, p[:end]...)
		b.isDataTruncated = true
		return end, nil
	} else {
		b.isDataTruncated = true
		return 0, nil
	}
}

File: v3/newrelic/secure_agent_test.go [hunk 2]
```diff
@@ -0,0 +1,460 @@
...
+			name:           "write exactly to limit",
+			initialBuf:     nil,
+			writeData:      []byte("1234567890"),
+			expectedN:      10,
+			expectedBufLen: 10,
+			expectedBuf:    "1234567890",
+			expectedTrunc:  false,
+		},
+		{
+			name:           "write partial when exceeding limit",
```
mvicknr, line 312: This case was failing before I made the changes to the `Write` method. It was returning:
 `expectedBuf: "12345abcd"`
 `expectedN: 4`
 `expectedTrunc: false`
so even though the limit was 10, we were only writing 9 and weren't marking it as truncated

v3/newrelic/secure_agent_test.go
```diff
@@ -0,0 +1,460 @@
...
+			initialBuf:     []byte("12345"),
+			writeData:      []byte("abcdefgh"),
+			expectedN:      5,
+			expectedBufLen: 10,
+			expectedBuf:    "12345abcde",
+			expectedTrunc:  true,
+		},
+		{
+			name:           "write when buffer at limit-1",
```
mvicknr, line 321: This case was failing before I made the changes to the Write method. It was returning:
`expectedBuf: "123456789"`
`expectedN: 0`
so even though the limit was 10, we were only writing 9

PR #1090, opened 2025-09-18
fix: Update Span Attribute db.Statement limit

cade-conklin:
### Issue
The Span Event Attribute `db.statement` is increasing from 2000 bytes to 4096 bytes
### Goals
Add logic to distributed tracing to account for span events up to 4096 bytes, and span events longer than 4096 bytes
### Implementation
- Added constant `MaxSpanAttributeDBStatementSize` to `tracing.go` to be set to 4096 and is subject to change if the max `db.statement` size changes
- Added function `truncateSpanAttribute` in `tracing.go` to sanitize any `db.statement` that is longer than 4096 bytes.  If longer, it will truncate the string to 4093 bytes (or the most amount of bytes that can be truncated if there are characters longer than a byte) with an ellipsis (`...`) appended to the end.
- Logic added in `tracing.go` is a part of the `endDataStoreSegment` function
- Changed `ParameterizedQuery` in `internal_span_events_test.go` to be longer than 4096 bytes.  Test passes with truncated version of string.
- Added unit tests in `tracing_test.go` to cover cases of implementation in `truncateSpanAttribute`
### How to test
- Use VsCode to run individual tests
- OR
- `make core-test TEST=newrelic`
- Tests should pass in Github actions as well


Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-464899

v3/newrelic/tracing_test.go
```diff
@@ -842,3 +842,84 @@ func TestBetterCAT_SetTraceAndTxnIDs(t *testing.T) {
...
+
+func Test_truncateSpanAttribute(t *testing.T) {
+	tests := []struct {
+		name string // description of this test case
+		// Named input parameters for target function.
+		value                 string
+		maxLengthWithEllipsis int
+		want                  string
+	}{
+		// TODO: Add test cases.
```
bduranleau-nr: Is this TODO still relevant?

v3/newrelic/tracing_test.go
```diff
@@ -842,3 +842,84 @@ func TestBetterCAT_SetTraceAndTxnIDs(t *testing.T) {
...
+			name:                  "Length of value is longer than maxLengthWithEllipsis and all characters larger than a byte",
+			value:                 "éééééé",
+			maxLengthWithEllipsis: 5,
+			want:                  "é...",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := truncateSpanAttribute(tt.value, tt.maxLengthWithEllipsis)
+			// TODO: update the condition below to compare got with tt.want.
```
bduranleau-nr: Is this TODO still relevant?

v3/newrelic/tracing_test.go
cade-conklin: It is not, good catch!

v3/newrelic/tracing_test.go
cade-conklin: Also not relevant, thank you!

v3/newrelic/tracing.go
```diff
@@ -843,17 +847,26 @@ func endDatastoreSegment(p endDatastoreParams) error {
...
+		evt.AgentAttributes.addString(SpanAttributeDBCollection, p.Collection, attributeValueLengthLimit)
 		p.TxnData.saveSpanEvent(evt)
 	}
 
 	return err
 }
 
+func truncateSpanAttribute(value string, maxLength int) string {
+
+	if len(value) > maxLength {
```
bduranleau-nr, line 863: So if `maxLength` is `4096`, and `len(value)` is something like `4097`, we would still throw away the last 3 bytes of data to replace with `...`, even though we're only one byte of data over the limit? In other words, intentionally toss 3 otherwise valid bytes of data?

bduranleau-nr: nvm, found the spec and this is definitely intended

v3/newrelic/tracing_test.go
```diff
@@ -842,3 +842,82 @@ func TestBetterCAT_SetTraceAndTxnIDs(t *testing.T) {
...
+		{
+			name:                  "Length of value is equal to maxLengthWithEllipsis and all characters larger than a byte",
+			value:                 "ééééé",
+			maxLengthWithEllipsis: 5,
+			want:                  "é...",
+		},
+		{
+			name:                  "Length of value is longer than maxLengthWithEllipsis and all characters larger than a byte",
+			value:                 "éééééé",
+			maxLengthWithEllipsis: 5,
```
bduranleau-nr, line 911: Could you add one test where the `maxLengthWithEllipsis` value is something like `6` (or any other nearby even number)? All the tests above are exercising odd numbers with multibyte string characters, I think it would be good to verify behavior for even number limits as well.

cade-conklin: Good catch.  This actually uncovered a bug that will truncate a character to only the max size - 3 bytes which could leave an incomplete character

cade-conklin: Follow up! Pushed up the changes and re-used the existing string concatenation that we have `stringLengthByteLimit` to properly concatenate multi-byte strings.  Odd thing though is when we call `addString` for the `db.statement` specifically, we actually call that function twice now (since it is called within `truncateSpanAttribute`.  This shouldn't be too much of a problem since it won't truncate twice because the length would be less than the max but I just think it's a little awkward.  Any ideas on a different approach?

bduranleau-nr: Documenting what we discussed here for posterity- A possible solution is to use the `key` string value passed to `addString` to gate DB vs non-DB behavior, which should eliminate the need for multiple truncate calls since the truncate logic can be handled directly in the `addString` function itself (for DB specifically)

v3/newrelic/tracing.go
```diff
@@ -17,6 +17,10 @@ import (
 	"github.com/newrelic/go-agent/v3/internal/logger"
 )
 
+const (
+	MaxSpanAttributeDBStatementSize = 4096 // int representing max size in bytes
```
bduranleau-nr: I think we want to relocate this to `limits.go`- that's where the other limit values (like `attributeValueLengthLimit`) already live

v3/newrelic/tracing.go
```diff
@@ -206,12 +210,16 @@ func (b boolJSONWriter) WriteJSON(buf *bytes.Buffer) {
 // value is a jsonWriter to allow for segment query parameters.
 type spanAttributeMap map[string]jsonWriter
 
+// addString writes the span attribute with any value over passed in limit
```
bduranleau-nr: This comment should be updated since we are no longer directly passing in the `limit` value

bduranleau-nr (approved):
Great work!

PR #1091, opened 2025-09-19
refactor: Size limits in distributed tracing

cade-conklin:
### Issue
Span attribute limits are 255 bytes which are not checked
### Goals
Check the span attribute limits
### Implementation
Whenever adding a string check the size limit using the existing string byte checker.  Special case and function made for DB.statement
### How to test
No tests have been specifically written yet


Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-465559

v3/newrelic/tracing.go
```diff
@@ -212,7 +212,13 @@ type spanAttributeMap map[string]jsonWriter
 
 func (m *spanAttributeMap) addString(key string, val string) {
 	if val != "" {
-		m.add(key, stringJSONWriter(val))
+		m.add(key, stringJSONWriter(stringLengthByteLimit(val, attributeValueLengthLimit)))
+	}
+}
+
+func (m *spanAttributeMap) addLargeString(key string, val string) {
```
mirackara: The name is fine, but we should add a comment here to explain why we are using this function vs. the traditional addString function

cade-conklin: Added the comments in!

v3/newrelic/tracing.go
```diff
@@ -816,7 +826,7 @@ func endDatastoreSegment(p endDatastoreParams) error {
 
 	if p.TxnData.TxnTrace.considerNode(end) {
 		attributes := end.agentAttributes.copy()
-		attributes.addString(SpanAttributeDBStatement, truncateSpanAttribute(p.ParameterizedQuery, MaxSpanAttributeDBStatementSize))
+		attributes.addLargeString(SpanAttributeDBStatement, p.ParameterizedQuery)
```
bduranleau-nr: Rather than create separate functions, why not unify all this unify all of this behavior under `addString` by allowing `addString` to take an integer argument for the length limit? This would allow for more dynamic setting of limits by the developer vs the hardcoded approach if, for example, we needed to limit another attribute to `2048` bytes. Something like this:
```
func (m *spanAttributeMap) addString(key string, val string, int limit)
```
For DB, you can pass the `truncateSpanAttribute` value still, since your finalized output from that will still match the byte limit.

cade-conklin: I like this idea! Just pushed up changes, let me know how they look!

mirackara (approved):

PR #1096, opened 2025-09-29
feat(config): Span event config controlled through env vars

cade-conklin:
## Links

[Example of python agent environment variables](https://docs.newrelic.com/docs/apm/agents/python-agent/configuration/python-agent-configuration/#environment-variables)
[Example of python agent setting Span Event env vars](https://github.com/newrelic/newrelic-python-agent/blob/7f3461d744dc0b63969a3fd7ab4189ce038192b9/newrelic/core/config.py#L846)

### Issue
The Go Agent does not control sending spans through environment variables
### Goals
Allow the Go Agent to control sending spans through environment variables 
### Implementation
- Added `MaxSamplesStored` field to `Config.SpanEvents`
- Added logic to assign values to `Config.SpanEvents.MaxSamplesStored` and `Config.SpanEvents.Enabled` via environment variables
### How to test
In progress...

### DIFF
--- a/v3/newrelic/config.go
+++ b/v3/newrelic/config.go
@@ -817,8 +817,8 @@ func (c Config) maxTxnEvents() int {
 	return configured
 }
 
-// maxSpan returns the configured maximum number of Span Events if it has been configured
-// and is less than the default maximum; otherwise it returns the default max.
+// maxSpan returns the configured maximum number of Span Events if it
+// is less than the default maximum; otherwise it returns the default max.
 func maxSpanEvents(configured int) int {
 	if configured < 0 || configured > internal.MaxSpanEvents {
 		return internal.MaxSpanEvents
--- a/v3/newrelic/config_options.go
+++ b/v3/newrelic/config_options.go
@@ -537,17 +537,15 @@ func configFromEnvironment(getenv func(string) string) ConfigOption {
 				}
 			}
 		}
-		assignInt := func(field *int, name string) {
+		assignInt := func(field *int, name string, fn func(configured int) int) {
 			if env := getenv(name); env != "" {
 				if i, err := strconv.Atoi(env); nil != err {
 					cfg.Error = fmt.Errorf("invalid %s value: %s", name, env)
 				} else {
-					switch name {
-					case "NEW_RELIC_SPAN_EVENTS_MAX_SAMPLES_STORED":
-						*field = maxSpanEvents(i)
-					default:
+					if fn == nil {
 						*field = i
-
+					} else {
+						*field = fn(i)
 					}
 				}
 			}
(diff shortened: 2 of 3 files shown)


v3/newrelic/config_options.go
```diff
@@ -577,6 +588,10 @@ func configFromEnvironment(getenv func(string) string) ConfigOption {
 		assignBool(&cfg.AIMonitoring.RecordContent.Enabled, "NEW_RELIC_AI_MONITORING_RECORD_CONTENT_ENABLED")
 		assignBool(&cfg.CustomInsightsEvents.CustomAttributesEnabled, "NEW_RELIC_APPLICATION_LOGGING_FORWARDING_CUSTOM_ATTRIBUTES_ENABLED")
 
+		// Span Event Env Variables
+		assignBool(&cfg.SpanEvents.Enabled, "NEW_RELIC_SPAN_EVENTS_ENABLED")
```
cade-conklin, line 600: @mirackara I added in this line to allow for control for enabling span events through env vars.  Should I add a function for this as well?

v3/newrelic/app_run.go
```diff
@@ -254,7 +254,7 @@ func (run *appRun) LoggingConfig() (config loggingConfig) {
 // which will be the default or the user's configured size (if any), but
 // may be capped to the maximum allowed by the collector.
 func (run *appRun) MaxSpanEvents() int {
-	return run.limit(internal.MaxSpanEvents, run.ptrSpanEvents)
+	return run.limit(run.Config.maxSpanEvents(), run.ptrSpanEvents)
```
bduranleau-nr: I took a look at [maxSpanEvents()](https://github.com/newrelic/go-agent/blob/1cd6e7b73662b2b3a2acbfb11b0b8084d9bf6a4b/v3/newrelic/app_run.go#L257), which calls [ptrSpanEvents()](https://github.com/newrelic/go-agent/blob/1cd6e7b73662b2b3a2acbfb11b0b8084d9bf6a4b/v3/newrelic/app_run.go#L222) to set the value, but it seems to differ from how other event types get handled.
Instead of setting the value in the [Reply.EventData.Limits](https://github.com/newrelic/go-agent/blob/1cd6e7b73662b2b3a2acbfb11b0b8084d9bf6a4b/v3/internal/connect_reply.go#L108) struct, we're setting a value in a unique [Reply.SpanEventHarvestConfig.HarvestLimit](https://github.com/newrelic/go-agent/blob/1cd6e7b73662b2b3a2acbfb11b0b8084d9bf6a4b/v3/internal/connect_reply.go#L109). Are we sure this is desired behavior? The EventData struct already has a field for [SpanEvents](https://github.com/newrelic/go-agent/blob/1cd6e7b73662b2b3a2acbfb11b0b8084d9bf6a4b/v3/internal/connect_reply.go#L125), so I would assume the value we want to set in `ptrSpanEvents()` would be this one instead.

bduranleau-nr: I think we may want to remove the `SpanEventHarvestConfig` struct from the `Reply` struct entirely and consolidate under the `EventData` struct. The only references I see to any of the `SpanEventHarvestConfig` struct are in the code referenced in the comment above, and some tests. I'm not sure it should be separate from the other events `harvest_limits` values.

bduranleau-nr: This comment can be ignored. After discussion, we determined that this is correct behavior based on the spec since span events cannot be included in the faster harvest cycle due to downsampling concerns. Additionally, this is not setting the json values, it is reading from the connect reply sent from the collector, not the connect response sent from the agent, which still should include `harvest_limits.span_event_data`.

v3/newrelic/config.go
```diff
@@ -812,6 +817,15 @@ func (c Config) maxTxnEvents() int {
 	return configured
 }
 
+// maxConfigEvents returns the configured maximum number of events if it has been configured
+// and is less than the default maximum; otherwise it returns the default max.
+func maxConfigEvents(configured int, max int) int {
+	if configured < 0 || configured > max {
```
bduranleau-nr: What should this behavior be if this value is zero? Should it be a defacto disable? This seems unlikely since we have a separate setting to enable/disable this behavior entirely. In some cases, it may mean the default configuration value, in others it may mean no upper bounds.

v3/newrelic/config.go
```diff
@@ -812,6 +817,15 @@ func (c Config) maxTxnEvents() int {
 	return configured
 }
 
+// maxConfigEvents returns the configured maximum number of events if it has been configured
+// and is less than the default maximum; otherwise it returns the default max.
+func maxConfigEvents(configured int, max int) int {
```
bduranleau-nr: This function name is a little bit misleading at a glance- it seems to imply that there is a `Config` event type that we are trying to return the max limit for, when in reality it's a helper function that is agnostic of Event type that we're using to check bounds on event types against customer configs. I think we may want to be careful generalizing this, and maybe it should just be specific to span events (i.e. `maxSpanEvents`) to avoid the need for a `max` arg (similar to how `maxTxnEvents` operate above and `maxCustomEvents` and `maxLogEvents` operate below). This allows us to definitively define what the `0` value behavior should be for span events.

v3/newrelic/config_options.go
```diff
@@ -544,16 +542,13 @@ func configFromEnvironment(getenv func(string) string) ConfigOption {
...
-					cfg.Error = fmt.Errorf("invalid %s value: %s", name, env)
-				} else {
-					*field = maxConfigEvents(i, max) // should send a warning back if the value is set to default
+					switch name {
+					case "NEW_RELIC_SPAN_EVENTS_MAX_SAMPLES_STORED":
+						*field = maxSpanEvents(i)
+					default:
+						*field = i
+
+					}
```
cade-conklin: @bduranleau-nr I added in this switch case to gate the checking of a valid value (only for span events) for env vars here.  In future PRs I can add other cases.  Is that a good compromise for checking env vars?

v3/newrelic/config.go
```diff
@@ -812,6 +817,15 @@ func (c Config) maxTxnEvents() int {
 	return configured
 }
 
+// maxSpan returns the configured maximum number of Span Events if it has been configured
+// and is less than the default maximum; otherwise it returns the default max.
+func maxSpanEvents(configured int) int {
```
bduranleau-nr, line 822: Should this be a method on the `Config` struct like the other `max*Events()` methods?

v3/newrelic/config_test.go
```diff
@@ -298,7 +299,7 @@ func TestCopyConfigReferenceFieldsPresent(t *testing.T) {
 				"span_event_data": %d
 			}
 		}
-	}]`, internal.MaxLogEvents, internal.MaxCustomEvents, internal.MaxSpanEvents, internal.MaxTxnEvents, internal.MaxCustomEvents, internal.MaxTxnEvents, internal.MaxSpanEvents))
+	}]`, internal.MaxLogEvents, internal.MaxCustomEvents, internal.MaxSpanEvents, internal.MaxSpanEvents, internal.MaxTxnEvents, internal.MaxCustomEvents, internal.MaxTxnEvents, internal.MaxSpanEvents))
```
bduranleau-nr, line 302: Do we want to reference `internal.MaxSpanEvents` 3 times here?

v3/newrelic/config.go
bduranleau-nr: Since we are changing the behavior to allow setting this config value by passing an arg, this approach is fine- other methods will updated in a future PR to follow this convention.

v3/newrelic/config_test.go
bduranleau-nr: Yup this is correct- DT == span events

Issue #983, opened 2024-11-01
Add integration for github.com/redis/rueidis


ankon:
## Summary

rueidis is a newer redis client for go, supporting streaming and automatic pipelining. In our testing we see improvements indeed, but right now there is no integration with the go-agent. 

## Desired Behaviour

Have a default integration, just like for go-redis/v{7,8,9}.

## Possible Solution

We implemented this, and it seems to work reasonably well.

[nrrueidis.go.txt](https://github.com/user-attachments/files/17598784/nrrueidis.go.txt)

Before converting this into a full-blown PR and adding boilerplate: WDYT?

iamemilio:
Just seeing this now, let us get back to you after the holidays

PR #1098, opened 2025-10-09
Add an integration for GitHub.com/redis/rueidis

ankon:
## Links

see #983: we have been using this code in production for quite some time now. 

## Details

The first commit fixes the dev-shell to work on Fedora Linux. It should be a no-op on other (non-SELinux-enabled) systems.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -74,7 +74,7 @@ services:
       <<: *mongo-env
 
     volumes:
-       - ${PWD}:/usr/src/app/go-agent
+       - ${PWD}:/usr/src/app/go-agent:rw,Z
     working_dir: /usr/src/app/go-agent
     stdin_open: true
     tty: true
--- a/v3/integrations/nrrueidis/nrrueidis.go
+++ b/v3/integrations/nrrueidis/nrrueidis.go
@@ -1,0 +1,137 @@
+package nrrueidis
+
+import (
+	"context"
+	"net"
+	"strings"
+	"time"
+
+	"github.com/newrelic/go-agent/v3/newrelic"
+	"github.com/redis/rueidis"
+	"github.com/redis/rueidis/rueidishook"
+)
+
+type contextKeyType struct{}
+
+var (
+	segmentContextKey = contextKeyType(struct{}{})
+)
+
+func multiOperation(typ string, cmds []rueidis.Completed) string {
+	operations := make([]string, 0, len(cmds))
+	for _, cmd := range cmds {
+		// XXX: Is it always true that there is only one command here?
+		operations = append(operations, cmd.Commands()[0])
+	}
+	return typ + ":" + strings.Join(operations, ",")
+}
+
+type hook struct {
+	segment newrelic.DatastoreSegment
+}
+
+func (h hook) before(ctx context.Context, operation string) context.Context {
+	txn := newrelic.FromContext(ctx)
+	if txn == nil {
+		return ctx
+	}
+	s := h.segment
+	s.Operation = operation
+	s.StartTime = txn.StartSegmentNow()
+	ctx = context.WithValue(ctx, segmentContextKey, &s)
+	return ctx
+}
+
+func (h hook) after(ctx context.Context) {
+	if segment, ok := ctx.Value(segmentContextKey).(interface{ End() }); ok {
+		segment.End()
+	}
+}
+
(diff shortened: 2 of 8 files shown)


ankon:
Hey there! Anything we can do to move this forward? It's a bit disappointing to not even get a reply from New Relic.

mirackara:
Hey @ankon 

Sorry about the delay. I'll review it now. Do you mind rebasing to our develop branch?

ankon:
hey @mirackara, I rebased it onto `master` now.

mirackara:
Hey @ankon, thanks! Do you mind also switching to our develop branch? otherwise this is looking good! I've also approved our workflows

docker-compose.yml
```diff
@@ -74,7 +74,7 @@ services:
       <<: *mongo-env
 
     volumes:
-       - ${PWD}:/usr/src/app/go-agent
+       - ${PWD}:/usr/src/app/go-agent:rw,Z
```
mirackara, line 77: (nit): volume mounts are read-write by default. we could simplify these lines to just use :Z

v3/integrations/nrrueidis/nrrueidis.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (h *hook) DoMulti(client rueidis.Client, ctx context.Context, multi ...rueidis.Completed) (resps []rueidis.RedisResult) {
+	ctx = h.before(ctx, multiOperation("pipeline", multi))
+	res := client.DoMulti(ctx, multi...)
+	h.after(ctx)
+	return res
+}
+
+func (h *hook) DoCache(client rueidis.Client, ctx context.Context, cmd rueidis.Cacheable, ttl time.Duration) (resp rueidis.RedisResult) {
+	// XXX: Is it always true that there is only one command here?
```
mirackara, line 67: In rueidis, calling .Commands() returns a slice of strings where index 0 is always the command name. I believe this is true and a safe way to get the command -- same for line 52. We can remove these comments

v3/integrations/nrrueidis/nrrueidis.go
```diff
@@ -0,0 +1,137 @@
...
+func (h *hook) Receive(client rueidis.Client, ctx context.Context, subscribe rueidis.Completed, fn func(msg rueidis.PubSubMessage)) (err error) {
+	ctx = h.before(ctx, subscribe.Commands()[0])
+	res := client.Receive(ctx, subscribe, fn)
+	h.after(ctx)
+	return res
+}
+
+func (h *hook) DoStream(client rueidis.Client, ctx context.Context, cmd rueidis.Completed) rueidis.RedisResultStream {
+	// XXX: Is it always true that there is only one command here?
+	// XXX: Should we mark this as "streaming"
```
mirackara, line 96: I think marking this as streaming is a good idea. we can clear these up too

v3/integrations/nrrueidis/nrrueidis.go
```diff
@@ -0,0 +1,137 @@
...
+func nrrueidisHook(opt rueidis.ClientOption) *hook {
+	h := &hook{}
+	h.segment.Product = newrelic.DatastoreRedis
+
+	if len(opt.InitAddress) == 0 {
+		return h
+	}
+
+	// Parse the first init address to get host and port.
+	// XXX: Does rueidis support unix sockets?
```
mirackara, line 128: APM dashboard won't be able to show where the Redis queries are going whenever a Unix socket is used. `net.SplitHostPort` only looks for port seperators. We could add a check before we call SplitHostPort here

```
addr := opt.InitAddress[0]

// Check if the address looks like a Unix domain socket path
if strings.HasPrefix(addr, "/") || strings.HasSuffix(addr, ".sock") {
    h.segment.Host = "localhost"
    h.segment.PortPathOrID = addr // New Relic allows paths/IDs in this field
} else {
    // Fall back to standard TCP parsing if it's not a Unix socket
    if host, port, err := net.SplitHostPort(addr); err == nil {
        ......
    }

```

docker-compose.yml
mirackara, line 1: Since rueidis requires a live Redis instance to execute integration tests against, should we spin up a standard redis image inside this docker-compose.yaml?

mirackara (changes requested):
If possible, having an example folder showing how you are using the integration would be great for others as a model. Thanks @ankon !

PR #1101, opened 2025-10-30
nrlambda: add support for MultiValueHeaders in event responses

### LINKED ISSUE #1072: nrlambda: Add support for APIGatewayProxyResponse.MultiValueHeaders
## Summary

Add support for the `MultiValueHeaders` property when extracting the headers from `events.APIGatewayProxyResponse` in nrlambda.

## Desired Behavior

Currently, nrlambda only extracts the `Headers` field.

https://github.com/newrelic/go-agent/blob/bc9e89fbd3fae7fa0d274f4eb817758bd118741b/v3/integrations/nrlambda/events.go#L98-L100

It should consider both the `Headers` and `MultiValueHeaders` fields, and merge them as described in [the AWS documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format).

> If you specify values for both headers and multiValueHeaders, API Gateway merges them into a single list. If the same key-value pair is specified in both, only the values from multiValueHeaders will appear in the merged list.

## Possible Solution

See above.

## Additional context

We are setting `MultiValueHeaders` not `Headers`, and thus currently none of the response headers are added as transaction attributes.

rittneje:
Fixes #1072.

## Links

n/a

## Details

Adds support for `MultiValueHeaders` in `APIGatewayProxyResponse` and `ALBTargetGroupResponse`.

### DIFF
--- a/v3/integrations/nrlambda/events.go
+++ b/v3/integrations/nrlambda/events.go
@@ -119,27 +119,43 @@ func eventWebRequest(event interface{}) *newrelic.WebRequest {
 func eventResponse(event interface{}) *response {
 	var code int
 	var headers map[string]string
+	var multiValueHeaders map[string][]string
 
 	switch r := event.(type) {
 	case events.APIGatewayProxyResponse:
 		code = r.StatusCode
 		headers = r.Headers
+		multiValueHeaders = r.MultiValueHeaders
 	case *events.APIGatewayProxyResponse:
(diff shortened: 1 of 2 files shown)


mirackara (approved):
LGTM! Thank you @rittneje

PR #1105, opened 2025-11-13
feat(nroci): First wrapper for OCI

cade-conklin:
### Issue
No integration for Oracle Cloud Infrastructure NoSQL SDK.  Need an initial outline to begin work
### Goals
Create a wrapper for the OCI Client that can be used in requests
### Implementation
Created wrapper for Client and Config.  Although there are no other fields in either wrapper, this is following a similarly implemented existing pattern that helps us future proof if we do need to add that (similar to the nropenai integration)
### Testing
n/a just a wrapper no functionality yet implemented

### DIFF
--- a/v3/integrations/nroci/go.mod
+++ b/v3/integrations/nroci/go.mod
@@ -1,0 +1,7 @@
+module github.com/newrelic/go-agent/v3/integrations/nroci
+
+go 1.24
+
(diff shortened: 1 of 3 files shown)


nr-swilloughby (approved):
LGTM, essentially this would be a drop-in replacement for the nosql library, satisfying the same interface the original does but wrapping all the calls to the underlying real library

PR #1106, opened 2025-11-19
feat(nroci): Client Do Request Implementation

cade-conklin:

### DIFF
--- a/v3/integrations/nroci/example/main.go
+++ b/v3/integrations/nroci/example/main.go
@@ -19,7 +19,7 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-	app.WaitForConnection(10 * time.Second)
+	app.WaitForConnection(10 * time.Second) // for short lived processes in apps
 	defer app.Shutdown(10 * time.Second)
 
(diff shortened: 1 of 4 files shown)


v3/integrations/nroci/example/main.go
```diff
@@ -0,0 +1,42 @@
...
+func main() {
+	app, err := newrelic.NewApplication(
+		newrelic.ConfigAppName("Basic NOSQLOCI App"),
+		newrelic.ConfigLicense(os.Getenv("NEW_RELIC_LICENSE_KEY")),
+		newrelic.ConfigDebugLogger(os.Stdout),
+	)
+	if err != nil {
+		panic(err)
+	}
+	app.WaitForConnection(10 * time.Second)
```
nr-swilloughby: Just to avoid misunderstandings when people follow our example code with their own implementations, we like to add a comment here saying that `WaitForConnection` is probably only needed for very short-lived programs or ones that have some specific need for the app to wait for instrumentation to be going before getting on with their work.

nr-swilloughby (approved):

PR #1117, opened 2026-01-21
feat(nroci): Initial implementation for oci-go-sdk

cade-conklin:
## Links
- [oci-go-sdk (nosql)](https://github.com/oracle/oci-go-sdk/tree/master/nosql)

### Integration
This PR contains an integration for the Oracle Cloud Infrastructure (OCI) Go SDK NoSQL package.
### Goals
- Create nroci package where this and future OCI integrations will live
- Create initial oci-go-sdk NoSQL integration
- Create example for oci-go-sdk NoSQL integration
### Implementation
The method used to integrate with the `oci-go-sdk` NoSQL package is to create a wrapper (of type `NoSQLClientWrapper` that has a field `Client` which implements `NoSQLClient`) and use those wrappers functions to send a `DatastoreSegment` when functions are called.  For example, you can get a client wrapper by calling our custom `NRNewNoSQLClientWithConfigurationProvider`:
```
clientWrapper, err := nroci.NRNewNoSQLClientWithConfigurationProvider(configProvider)
```
The `clientWrapper` contains the client from the `oci-go-sdk` in the field `clientWrapper.Client`.  Additionally, `configWrapper` has the same methods that the client implements from the interface `NoSQLClient` although it doesn't explicitly implement this interface.  This allows us to call our custom functions in the same way that the user would call the `oci-go-sdk` functions.  For example when making a query it is just:
```
queryRes, err := clientWrapper.Query(ctx, queryReq) // clientWrapper.Query(ctx context.Context, queryReq nosql.QueryRequest) takes in same parameters as nosql.Query()
```
Once `clientWrapper.Query(ctx, queryReq)` (or any of the functions belonging to `clientWrapper`) is called, it will return another call to `executeWithDatastoreSegmentOCI(ctx context.Context, req QueryRequest, fn func() (*QueryResponse, error), operation string)`.  The function that is passed into `executeWithDatastoreSegmentOCI` contains the logic to extract the fields for a `DatastoreSegment`,  start a `DatastoreSegment`, and call the original `oci-go-sdk` function (in this case `nosql.Query()` which is contained in the `clientWrapper.Client.Query()`.  

Also note that `executeWithDatastoreSegmentOCI` is defined to be type a generic function.  Here is what the definition looks like:
```
func executeWithDatastoreSegmentOCI[T any, R any](
	ctx context.Context,
	req R,
	fn func() (T, error),
	operation string,
) (*T, error) 
```
`T` is the request type and `R` is the response type.  The reasons we do this is because each different type of request from the `oci-go-sdk` has a different type.  In order to handle the different fields that get extracted from the different request types, we call `extractRequestFieldsOCI(req any)` and use a switch on the request type.  The default for any of the values that come back are an empty string.

Here is an example of how `executeWithDatastoreSegmentOCI` would be called from `clientWrapper.Query()`:
```
executeWithDatastoreSegmentOCI(ctx, req, func() (nosql.QueryResponse, error) {
		return cw.Client.Query(ctx, req)
	}, "Query")
```
Notice that the function calls `cw.Client.Query(ctx, req)`.  This is the original client's `Query` function.  Additionally, `"Query"` is passed in as a string to note the operation for the `DatastoreSegment`.  Each `clientWrapper` function has a different operation that is the same as the function name because of the uniqueness of the OCI NoSQL DB.  It is a DB that has no real database name, instead separated by compartment.  Additionally, it is queried with SQL, but at the same time, doesn't have real inserts, updates, deletes as apart of its language.

#### Unit test
```
go test ./...
```
#### Smoke test
```
# starting from go-agent
> cd v3/integrations/nroci
> go mod tidy
> cd example/oci-go-sdk
> go run main.go
```
Navigate to your New Relic account and find `Basic NOSQLOCI App`.  Check DB transactions there or in NRQL.

### DIFF
--- a/v3/integrations/nroci/go.mod
+++ b/v3/integrations/nroci/go.mod
@@ -1,4 +1,4 @@
-module github.com/newrelic/go-agent/v3/integrations/norci
+module github.com/newrelic/go-agent/v3/integrations/nroci
 
 go 1.24
 


v3/integrations/nroci/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/newrelic/go-agent/v3/integrations/norci
```
nr-swilloughby: shouldn't that be nroci?

nr-swilloughby (approved):
you're quick :)

PR #1120, opened 2026-02-12
feat(nrawssdk): Entity Relationship AWS update

cade-conklin:
#### Entity Relationships for Elastic/Open Search
- Added additional Span attribute for Opensearch/elastic search endpoint 
#### Entity Relationships, AWS account information config
-  Accept `NEW_RELIC_CLOUD_AWS_ACCOUNT_ID` as env var
- `Config.CloudAWS.AccountID` and `Config.CloudAWS.AccountDecoding.Enabled` configurable via config options (`ConfigCloudAWSAccountID` and `ConfigCloudAWSAccountDecodingEnabled`
- `Config.CloudAWS.AccountDecoding.Enabled` enabled by default
- Function `validateAWSAccountID` added to check when accountID is set in env or by config options
- `InitializeMiddleware` function added to `nrawssdk-v2`.  Contains logic to retrieve credentials from awsConfig and use them to resolve accountID (with Config setting given priority) and then add existing middlewares to be used in AWS requests
- `AppendMiddlewares` given a note that it is deprecated.  It will not add the accountID as a span attribute
- `accountID` added as a span attribute in `serializeMiddlewares`
- Example updated to reflect new changes in `InitializeMiddleware`.  It is no longer passed in as an optional function, but rather after the awsConfig is resolved.  Still an option to pass `InitializeMiddleware` as an optional function as detailed in comments for `InitializeMiddleware` description and in comments in example
- Interface added for accountID resolver to make mocking for tests easier
#### Entity Relationships, decoding AWS account ID
- `accountID` added as field in `nrMiddleware` struct
- function to decode the access key into an account ID as detailed [here](https://hackingthe.cloud/aws/enumeration/get-account-id-from-keys/)

### DIFF
--- a/v3/integrations/nrawssdk-v2/nrawssdk.go
+++ b/v3/integrations/nrawssdk-v2/nrawssdk.go
@@ -135,10 +135,9 @@ func (m nrMiddleware) deserializeMiddleware(stack *smithymiddle.Stack) error {
 				}
 			}
 			if serviceName == "OpenSearch" || serviceName == "opensearch" {
-				integrationsupport.AddAgentSpanAttribute(txn, newrelic.AttributeAWSElastSearchDomainEndpoint, httpRequest.URL.String()) // this way I don't have to pull it out of context
+				integrationsupport.AddAgentSpanAttribute(txn, newrelic.AttributeAWSElastSearchDomainEndpoint, httpRequest.URL.String())
 			}
-			// Set additional span attributes
-			integrationsupport.AddAgentSpanAttribute(txn, newrelic.AttributeCloudAccountID, accountID) // setting account ID here, why do we only do this if it is an SQS service?
+			integrationsupport.AddAgentSpanAttribute(txn, newrelic.AttributeCloudAccountID, accountID)
 
 			integrationsupport.AddAgentSpanAttribute(txn,
 				newrelic.AttributeResponseCode, strconv.Itoa(response.StatusCode))


v3/newrelic/config.go
```diff
@@ -93,6 +93,10 @@ type Config struct {
 		MaxSamplesStored int
 	}
 
+	CloudAWS struct {
```
mirackara, line 99: We should notate why we have this config option with a good comment

v3/integrations/nrawssdk-v2/nrawssdk.go
```diff
@@ -222,6 +236,28 @@ func AppendMiddlewares(apiOptions *[]func(*smithymiddle.Stack) error, txn *newre
 	*apiOptions = append(*apiOptions, m.serializeMiddleware)
 }
 
+func NRAppendMiddlewares(apiOptions *[]func(*smithymiddle.Stack) error, ctx context.Context, awsConfig aws.Config) {
```
mirackara: Are we differentiating our AppendMiddlewares vs. NRAppendMiddlewares enough here? What purpose does the older one still serve & is it something we can remove/combine?

cade-conklin: I should leave more comments here, but I put in a different function since it takes in different parameters!  I can discuss the purpose/parameters passed into this function since I think its a bit clunky as it stands

mirackara: Note from mobbing: can we have the old function just call this new function?

mirackara: Note from mobbing: can we just pass in one awsConfig aws.Config instead of aws.Config.apiOptions + aws.Config?

cade-conklin: @mirackara I am keeping it the same way because we still keep the option to inject our middleware into specific client requests.  The mismatching typing for the structs contains `apiOptions` from the client and the `awsConfig` prevent a good way to just pass one type in

v3/integrations/nrawssdk-v2/nrawssdk.go
```diff
@@ -278,3 +308,53 @@ func dynamoDBInputFromMiddlewareInput(in middleware.InitializeInput) dynamodbInp
...
+	}
+	var bigEndian uint64
+	for i := 0; i < 6; i++ {
+		bigEndian = bigEndian << 8      // shift 8 bits left.  Most significant byte read in first (decoded[i])
+		bigEndian |= uint64(decoded[i]) // apply OR for current byte
+	}
+
+	mask := uint64(0x7fffffffff80)
+
+	num := (bigEndian & mask) >> 7 // apply mask and get rid of last 7 bytes from mask
```
nr-swilloughby: bits?

cade-conklin: Yes, thank you!

v3/integrations/nrawssdk-v2/nrawssdk.go
```diff
@@ -125,7 +134,12 @@ func (m nrMiddleware) deserializeMiddleware(stack *smithymiddle.Stack) error {
 
 				}
 			}
+			if serviceName == "OpenSearch" || serviceName == "opensearch" {
+				integrationsupport.AddAgentSpanAttribute(txn, newrelic.AttributeAWSElastSearchDomainEndpoint, httpRequest.URL.String()) // this way I don't have to pull it out of context
```
mirackara: remove extraneous comment

v3/integrations/nrawssdk-v2/nrawssdk.go
```diff
@@ -125,7 +134,12 @@ func (m nrMiddleware) deserializeMiddleware(stack *smithymiddle.Stack) error {
...
+			}
 			// Set additional span attributes
+			integrationsupport.AddAgentSpanAttribute(txn, newrelic.AttributeCloudAccountID, accountID) // setting account ID here, why do we only do this if it is an SQS service?
+
```
mirackara, line 141: remove extraneous comment

PR #1123, opened 2026-02-25
feat(nrlambda): Add attribute for aws.lambda.eventSource.eventType

cade-conklin:
### Issue
`aws.lambda.eventSource.eventType` is not added as an agent attribute
### Goals
Add the above event type as an agent attribute, depending on the event, for all supported events in the lambda aws sdk
### Implementation
The event type is added similarly to the `eventSourceArn` with a switch statement.  It is added as an agent attribute in the same area
### How To test
I have local lambdas set up to test different event types.


nr-swilloughby (approved):
LGTM

Issue #1121, opened 2026-02-17
SQL parser fails to recognize WITH clause (Common Table Expressions)


andreameneghello:
The SQL parser doesn't recognize queries starting with `WITH` (CTEs). The `sqlOperations` map in `v3/internal/sqlparse/sqlparse.go` is missing an entry for `"with"`, causing these queries to not be tracked in New Relic.

## Steps to Reproduce

1. Use `nrpgx5.NewTracer()` with the PostgreSQL driver
2. Execute a CTE query:
```sql
WITH temp_results AS (
  SELECT id, name FROM users WHERE active = true
)
SELECT * FROM temp_results
```
3. Query is not properly parsed/tracked

## Expected Behavior

CTE queries should be recognized and tracked like other SQL operations.

## Your Environment

- **Go Agent:** v3 (latest)
- **Integration:** `nrpgx5`
- **Database:** PostgreSQL

## Root Cause

The `sqlOperations` map in `v3/internal/sqlparse/sqlparse.go` is missing:
```go
"with": nil,
```

## Additional context

**Proposed Fix:**

Add `"with": nil` to the `sqlOperations` map, following the same pattern as `"call"`, `"create"`, `"drop"`, etc.

CTEs are a standard SQL feature (SQL:1999) widely used in PostgreSQL. Users shouldn't need custom parsers for this common operation.

**Current workaround:** Custom parser function with `"with"` added to the map.

---

PR #1124, opened 2026-02-26
feat: add support for 'WITH' clause in SQL parsing

mirackara:
Add support for "WITH" clause in SQL parsing

Addresses https://github.com/newrelic/go-agent/issues/1121

### DIFF
--- a/v3/newrelic/sqlparse/sqlparse_test.go
+++ b/v3/newrelic/sqlparse/sqlparse_test.go
@@ -198,10 +198,11 @@ func TestExtractTable(t *testing.T) {
 }
 func TestParseSQLWith(t *testing.T) {
 	for _, tc := range []sqlTestcase{
-		{Input: "WITH cte AS (SELECT * FROM users) SELECT * FROM cte", Operation: "with", Table: ""},
-		{Input: "WITH RECURSIVE cte AS (SELECT id FROM employees) SELECT * FROM cte", Operation: "with", Table: ""},
-		{Input: "with temp AS (SELECT id FROM foo) SELECT * FROM temp", Operation: "with", Table: ""},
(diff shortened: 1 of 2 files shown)


v3/newrelic/sqlparse/sqlparse_test.go
```diff
@@ -196,3 +196,13 @@ func TestExtractTable(t *testing.T) {
 		}
 	}
 }
+func TestParseSQLWith(t *testing.T) {
```
cade-conklin, line 199: Add a test case using WHERE as not the first word to be parsed and check if proper operation is picked up.  Could be as a subquery added to `TestParseSQLSubQuery` or add a test checking WITH and operations where `Input` contains multiple queries i.e.
```
{Input: "Select * FROM cte; WITH cte AS (SELECT * FROM users) SELECT * FROM cte;", Operation: "select", Table: "cte"}
```

v3/newrelic/sqlparse/sqlparse_test.go
```diff
@@ -196,3 +196,13 @@ func TestExtractTable(t *testing.T) {
...
+	for _, tc := range []sqlTestcase{
+		{Input: "WITH cte AS (SELECT * FROM users) SELECT * FROM cte", Operation: "with", Table: ""},
```
cade-conklin: Does Table (`newrelic.DatastoreSegment.Collection`) not get populated in this case?

mirackara: Your right, we need a regex for it

cade-conklin (approved):
Approved with one nit: maybe we should put in the comments of `ParseQuery` what type of queries we support or how it works.  For example, we parse the first word and then identify queries based on that keyword, but if the query starts with something like "FROM" it won't parse anything.

PR #1131, opened 2026-03-18
refactor(nrgocql): Create segment during exec

cade-conklin:

### DIFF
--- a/v3/integrations/nrgocqlx/nrgocqlx.go
+++ b/v3/integrations/nrgocqlx/nrgocqlx.go
@@ -20,7 +20,6 @@ package nrgocqlx
 import (
 	"context"
 	"strconv"
-	"strings"
 
 	gocql "github.com/gocql/gocql"
 	"github.com/newrelic/go-agent/v3/internal"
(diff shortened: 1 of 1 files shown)


v3/integrations/nrgocql/nrgocql.go
```diff
@@ -36,6 +37,85 @@ type queryObserver[T any] struct {
...
+	}
+
+	// start datastore segment
+	sgmt := &newrelic.DatastoreSegment{
+		StartTime: txn.StartSegmentNow(),
+		Product:   newrelic.DatastoreCassandra,
+	}
+
+	// do I need newrelic security agent?
+
```
mirackara: we should add secure agent here

`	if newrelic.IsSecurityAgentPresent() {
		sgmt.SetSecureAgentEvent(secureAgentEvent)
	}`

v3/integrations/nrgocql/nrgocql.go
```diff
@@ -36,6 +37,85 @@ type queryObserver[T any] struct {
...
```
mirackara, line 87: same as above ^

PR #1163, opened 2026-04-29
feat: Add New Relic instrumentation for Anthropic SDK integration

mirackara:
Implements the nranthropic integration package, which wraps the [Anthropic Go SDK](https://github.com/anthropics/anthropic-sdk-go) with New Relic AI Monitoring instrumentation

### DIFF
--- a/v3/integrations/nranthropic/nranthropic.go
+++ b/v3/integrations/nranthropic/nranthropic.go
@@ -1,0 +1,241 @@
+// Package nranthropic provides New Relic instrumentation for the Anthropic Go
+// SDK (github.com/anthropics/anthropic-sdk-go).
+//
+// Wrap your Anthropic client with NewClient, then use nrClient.Messages.New in
+// place of client.Messages.New to automatically record LlmChatCompletionSummary
+// and LlmChatCompletionMessage custom events and a segment under the active
+// transaction — mirroring the Python agent's mlmodel_anthropic instrumentation.
+//
+// The New Relic transaction must be present in the context (via
+// newrelic.NewContext) for instrumentation to activate. AI monitoring must also
+// be enabled on the application (newrelic.ConfigAIMonitoringEnabled(true)).
+package nranthropic
+
+import (
+	"context"
+	"fmt"
+	"runtime/debug"
+	"strings"
+	"time"
+
+	anthropic "github.com/anthropics/anthropic-sdk-go"
+	"github.com/anthropics/anthropic-sdk-go/option"
+	"github.com/google/uuid"
+	"github.com/newrelic/go-agent/v3/internal"
+	"github.com/newrelic/go-agent/v3/newrelic"
+	"github.com/newrelic/go-agent/v3/newrelic/integrationsupport"
+)
+
+// Version is the current version of the nranthropic integration.
+const Version = "0.1.0"
+
+func init() {
+	info, ok := debug.ReadBuildInfo()
+	if info != nil && ok {
+		for _, module := range info.Deps {
+			if module != nil && strings.Contains(module.Path, "anthropic-sdk-go") {
+				internal.TrackUsage("Go", "ML", "Anthropic", module.Version)
+				return
+			}
+		}
+	}
+	internal.TrackUsage("Go", "ML", "Anthropic", "unknown")
+}
+
+// NRClient wraps an Anthropic client with New Relic instrumentation.
+type NRClient struct {
+	Client           *anthropic.Client
+	Messages         NRMessageService
+	customAttributes map[string]interface{}
+}
+
+// NRMessageService wraps the Anthropic MessageService with instrumentation.
+type NRMessageService struct {
+	client *anthropic.Client
+	app    *newrelic.Application
+	nrc    *NRClient
+}
+
+// NewClient creates an NRClient wrapping the given Anthropic client.
+func NewClient(app *newrelic.Application, client *anthropic.Client) *NRClient {
+	nrc := &NRClient{Client: client}
+	nrc.Messages = NRMessageService{client: client, app: app, nrc: nrc}
+	return nrc
+}
+
+// AddCustomAttributes attaches llm.* prefixed key-value pairs to all LLM events
+// recorded by this client.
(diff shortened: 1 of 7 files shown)


v3/integrations/nranthropic/nranthropic.go
```diff
@@ -0,0 +1,464 @@
...
+		ctx = newrelic.NewContext(ctx, txn)
+	}
+
+	integrationsupport.AddAgentAttribute(txn, "llm", "", true)
+
+	completionID := uuid.New().String()
+	spanID := txn.GetTraceMetadata().SpanID
+	traceID := txn.GetTraceMetadata().TraceID
+
+	seg := txn.StartSegment("Llm/completion/Anthropic/MessageNew")
```
cade-conklin: nit: The spec calls for this to be instrumented as the function name: 
```
function_name: Name of instrumented function (ex: invoke_model, create)
```
It should technically be `seg := txn.StartSegment("Llm/completion/Anthropic/New")`

v3/integrations/nranthropic/nranthropic.go
```diff
@@ -0,0 +1,464 @@
...
+		start:            time.Now(),
+	}
+
+	if !cfg.AIMonitoring.Enabled || !cfg.AIMonitoring.Streaming.Enabled {
+		w.stream = s.messages.NewStreaming(ctx, params, opts...)
+		return w
+	}
+
+	integrationsupport.AddAgentAttribute(txn, "llm", "", true)
+	seg := txn.StartSegment("Llm/completion/Anthropic/MessageNewStreaming")
```
cade-conklin: nit: The spec calls for this to be instrumented as the function name: 
```
function_name: Name of instrumented function (ex: invoke_model, create)
```
It should technically be `seg := txn.StartSegment("Llm/completion/Anthropic/NewStreaming")`

v3/integrations/nranthropic/nranthropic.go
```diff
@@ -0,0 +1,464 @@
...
+}
+
+// NewStreaming wraps client.Messages.NewStreaming with New Relic instrumentation.
+// The returned wrapper exposes Next/Current/Err/Close matching the underlying
+// ssestream.Stream API. Call Close() after the stream is consumed to record NR events.
+func (s *NRMessageService) NewStreaming(ctx context.Context, params anthropic.MessageNewParams, opts ...option.RequestOption) *NRMessageStreamWrapper {
+	cfg, _ := s.app.Config()
+
+	if !cfg.AIMonitoring.Streaming.Enabled {
+		if reportStreamingDisabled != nil {
```
cade-conklin, line 278: Question: Can this logic be moved to within the if statement on line 302? We check if streaming is enabled there as well and it doesn't seem like `reportStreamingDisabled` will cause the function to return early.  Alternatively, do we want to keep this logic here and force an early return with an uninitialized `NRStreamWrapper`?  If their streaming isn't enabled it would make sense to not start a transaction.

cade-conklin: Adding to this comment, I think we need to change the logic so that we don't begin a transaction if streaming or ai monitoring is disabled.

v3/integrations/nranthropic/nranthropic.go
```diff
@@ -0,0 +1,464 @@
...
+// Err returns any error encountered during streaming.
+func (w *NRMessageStreamWrapper) Err() error {
+	return w.stream.Err()
+}
+
+// Close records LlmChatCompletionSummary and LlmChatCompletionMessage events,
+// ends the segment, ends the transaction if it was started by this wrapper,
+// and closes the underlying stream.
+func (w *NRMessageStreamWrapper) Close() error {
+	err := w.stream.Close()
```
cade-conklin: Refactor: Change the logic to return the stream at the end of Close() and make recording a custom event itts own function.  It is awkward to close the stream, but then hang onto the error until its returned at the end.  This would allow to just return the stream close at the end, only check for the txn once, and allow to mock logic for recording custom events in test.  Example:
```
func (w *NRMessageStreamWrapper) Close() error {
	w.recordCustomEvent()
	if w.txnOwned {
		w.txn.End()
	}
	return w.stream.Close()
}

func (w *NRMessageStreamWrapper) recordCustomEvent() {
	cfg, _ := w.app.Config()
	if !cfg.AIMonitoring.Enabled || !cfg.AIMonitoring.Streaming.Enabled {
		return
	}
	...
	...
	// rest of custom event logic
}
```

v3/integrations/nranthropic/nranthropic_test.go
cade-conklin, line 1: Kudos: There are a lot of tests for `New` and `NewStreaming`

cade-conklin, line 1: Suggestion: Add unit tests for smaller functions.  I.e. test the logic for `recordSummary`, `recordMessages`, `extractParamText`, `extractResponseText`, `NRMessageStreamWrapper.Next`, `NRMessageStreamWrapper.Close` (after refactor), `appendCustomAttributes`, and `NRClient.AddCustomAttributes`.

v3/integrations/nranthropic/nranthropic.go
```diff
@@ -0,0 +1,464 @@
...
+}
+
+// NRMessageStreamWrapper wraps an Anthropic SSE stream with New Relic instrumentation.
+// Call Next() to advance the stream, Current() to read the current event, Err() to
+// check for errors, and Close() to flush NR events and release resources.
+type NRMessageStreamWrapper struct {
+	stream           *ssestream.Stream[anthropic.MessageStreamEventUnion]
+	app              *newrelic.Application
+	txn              *newrelic.Transaction
+	txnOwned         bool
```
cade-conklin, line 257: Kudos: This is a really clever way to contain a transaction if one wasn't started in accordance with the spec

cade-conklin (changes requested):

v3/integrations/nranthropic/nranthropic.go
mirackara: Yup good catch. If it's disabled we should not start a txn at all

cade-conklin (approved):
Well done! :shipit:

PR #1189, opened 2026-07-20
(Gemini) go-genai integration support

mirackara:
## Details

- Added Support for `go-genai`(Gemini integration)
  - Support for `GenerateContent` / `GenerateContentStream`
  - `nrgemini.GenerateContent()` 
    - Calls go-genai's `GenerateContent()` 
    - After a response, record `LlmChatCompletionSummary` and `LlmChatCompletionMessage` events
- Preserved full-length LLM event attributes being sent via custom events

### DIFF
--- a/v3/integrations/nrgemini/nrgemini.go
+++ b/v3/integrations/nrgemini/nrgemini.go
@@ -1,0 +1,543 @@
+// Package nrgemini provides New Relic instrumentation for the Google GenAI Go
+// SDK (google.golang.org/genai).
+//
+// Wrap your GenAI client with NewClient, then use nrClient.Models.GenerateContent
+// in place of client.Models.GenerateContent to automatically record
+// LlmChatCompletionSummary and LlmChatCompletionMessage custom events and a
+// segment under the active transaction — mirroring the Python agent's
+// mlmodel_gemini instrumentation.
+//
+// The New Relic transaction must be present in the context (via
+// newrelic.NewContext) for instrumentation to activate. AI monitoring must also
+// be enabled on the application (newrelic.ConfigAIMonitoringEnabled(true)).
+package nrgemini
+
+import (
+	"context"
+	"fmt"
+	"iter"
+	"runtime/debug"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/newrelic/go-agent/v3/internal"
+	"github.com/newrelic/go-agent/v3/newrelic"
+	"github.com/newrelic/go-agent/v3/newrelic/integrationsupport"
+	"google.golang.org/genai"
+)
+
+var reportStreamingDisabled func()
+
+func init() {
+	reportStreamingDisabled = sync.OnceFunc(func() {
+		internal.TrackUsage("Go", "ML", "Streaming", "Disabled")
+	})
+	info, ok := debug.ReadBuildInfo()
+	if info != nil && ok {
+		for _, module := range info.Deps {
+			if module != nil && strings.Contains(module.Path, "google.golang.org/genai") {
+				internal.TrackUsage("Go", "ML", "Gemini", module.Version)
+				return
+			}
+		}
+	}
+	internal.TrackUsage("Go", "ML", "Gemini", "unknown")
+}
+
+// NRClient wraps a GenAI client with New Relic instrumentation.
+type NRClient struct {
+	Client *genai.Client
+	Models NRModelsService
+}
+
+// NRModelsService wraps the GenAI Models service with instrumentation.
+type NRModelsService struct {
+	models           *genai.Models
+	app              *newrelic.Application
+	customAttributes map[string]interface{}
+}
+
+// NewClient creates an NRClient wrapping a new GenAI client initialized with cfg.
+func NewClient(app *newrelic.Application, ctx context.Context, cfg *genai.ClientConfig) (*NRClient, error) {
+	client, err := genai.NewClient(ctx, cfg)
+	if err != nil {
+		return nil, err
+	}
+	nrc := &NRClient{Client: client}
+	nrc.Models = NRModelsService{
+		models:           client.Models,
+		app:              app,
+		customAttributes: make(map[string]interface{}),
+	}
+	return nrc, nil
+}
+
+// AddCustomAttributes attaches llm.* prefixed key-value pairs to all LLM events
+// recorded by this client.
+func (c *NRClient) AddCustomAttributes(attrs map[string]interface{}) {
+	for k, v := range attrs {
+		if strings.HasPrefix(k, "llm.") {
+			c.Models.customAttributes[k] = v
+		}
+	}
+}
+
+// GenerateContent wraps client.Models.GenerateContent with New Relic instrumentation.
+//
+// If ctx carries a New Relic transaction (via newrelic.NewContext) that
+// transaction is used and its lifecycle is left to the caller. If no
+// transaction is present a new one named "GeminiGenerateContent" is started and
+// ended automatically. AI monitoring must be enabled on the application
+// (newrelic.ConfigAIMonitoringEnabled(true)); otherwise the call is forwarded
+// to the underlying SDK without instrumentation.
+func (s *NRModelsService) GenerateContent(ctx context.Context, model string, contents []*genai.Content, config *genai.GenerateContentConfig) (*genai.GenerateContentResponse, error) {
+	cfg, _ := s.app.Config()
+	if !cfg.AIMonitoring.Enabled {
+		return s.models.GenerateContent(ctx, model, contents, config)
+	}
+
+	txn := newrelic.FromContext(ctx)
+	if txn == nil {
+		txn = s.app.StartTransaction("GeminiGenerateContent")
+		defer txn.End()
+		ctx = newrelic.NewContext(ctx, txn)
+	}
+
+	integrationsupport.AddAgentAttribute(txn, "llm", "", true)
+
+	completionID := uuid.New().String()
+	spanID := txn.GetTraceMetadata().SpanID
+	traceID := txn.GetTraceMetadata().TraceID
+
+	seg := txn.StartSegment("Llm/completion/Gemini/GenerateContent")
+	start := time.Now()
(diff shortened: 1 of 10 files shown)


v3/integrations/nrgemini/nrgemini.go
```diff
@@ -0,0 +1,543 @@
...
+	if !cfg.AIMonitoring.Enabled {
+		return s.models.GenerateContent(ctx, model, contents, config)
+	}
+
+	txn := newrelic.FromContext(ctx)
+	if txn == nil {
+		txn = s.app.StartTransaction("GeminiGenerateContent")
+		defer txn.End()
+		ctx = newrelic.NewContext(ctx, txn)
+	}
```
cade-conklin, line 106: issue: Should we be starting a transaction if one doesn't exist in context? I believe that should be on the customer. We should return and run the original `GenerateContent` here

v3/integrations/nrgemini/nrgemini.go
```diff
@@ -0,0 +1,543 @@
...
+		data["error"] = true
+		data["response.number_of_messages"] = len(contents)
+	} else if resp != nil {
+		if resp.ModelVersion != "" {
+			data["response.model"] = resp.ModelVersion
+		}
+		if len(resp.Candidates) > 0 && resp.Candidates[0].FinishReason != "" {
+			data["response.choices.finish_reason"] = string(resp.Candidates[0].FinishReason)
+		}
+		data["response.number_of_messages"] = len(contents) + 1
```
cade-conklin, line 168: question: Why add 1 here?

v3/integrations/nrgemini/nrgemini.go
```diff
@@ -0,0 +1,543 @@
...
+		return resp, err
+	}
+
+	s.recordSummary(completionID, spanID, traceID, model, config, contents, resp, duration, false)
+	s.recordMessages(completionID, spanID, traceID, model, contents, resp)
+	return resp, nil
+}
+
+func (s *NRModelsService) recordSummary(completionID, spanID, traceID, model string, config *genai.GenerateContentConfig, contents []*genai.Content, resp *genai.GenerateContentResponse, duration int64, isError bool) {
+	data := map[string]interface{}{
```
cade-conklin, line 139: issue: `request_id` doesn't look like it is checked for.  That should be available in the response headers.  Check if full response is not nil then check for response headers.  Something like:
```
	resp.SDKHTTPResponse.Headers.Get("request_id")
```

cade-conklin: issue: similarly check headers for `response.organization`

v3/integrations/nrgemini/nrgemini.go
cade-conklin, line 168: question: Are these two things mutually exculsive? For instance, if there is an error, is ther response always `nil`?

v3/integrations/nrgemini/nrgemini.go
cade-conklin: issue: Missing `response.usage.prompt_tokens`, `response.usage.completion_tokens`, and `response.usage.total_tokens`.  Are we not including those for a reason?

cade-conklin: issue: Are we checking for `response.headers.<vendor_specific_headers>` to add to the custom event?

v3/integrations/nrgemini/nrgemini.go
```diff
@@ -0,0 +1,543 @@
...
+		data["token_count"] = tokens
+	}
+	s.appendCustomAttrs(data)
+	s.app.RecordCustomEvent("LlmChatCompletionMessage", data)
+}
+
+func (s *NRModelsService) appendCustomAttrs(data map[string]interface{}) {
+	for k, v := range s.customAttributes {
+		data[k] = v
+	}
```
cade-conklin, line 244: nit: Replace with `	maps.Copy(data, s.customAttributes)
`

v3/integrations/nrgemini/nrgemini.go
```diff
@@ -0,0 +1,543 @@
...
+
+	for i, c := range contents {
+		text := extractContentText(c)
+		role := c.Role
+		if role == "" {
+			role = genai.RoleUser
+		}
+		data := map[string]interface{}{
+			"id":             uuid.New().String(),
```
cade-conklin, line 189: issue: Should check if a `resp.ResponseID` exists with the sequence beginning at index 0.  UUID is fallback

v3/integrations/nrgemini/nrgemini.go
```diff
@@ -0,0 +1,543 @@
...
+	}
+
+	responseText := extractResponseText(resp)
+	responseSeq := len(contents)
+	responseRole := genai.RoleModel
+	if len(resp.Candidates) > 0 && resp.Candidates[0].Content != nil && resp.Candidates[0].Content.Role != "" {
+		responseRole = resp.Candidates[0].Content.Role
+	}
+	data := map[string]interface{}{
+		"id":             fmt.Sprintf("%s-%d", resp.ResponseID, responseSeq),
```
cade-conklin, line 220: issue: Needs to check if that response id exists and if not add a UUID

v3/integrations/nrgemini/nrgemini.go
```diff
@@ -0,0 +1,543 @@
...
+
+	for i, c := range contents {
+		text := extractContentText(c)
+		role := c.Role
+		if role == "" {
+			role = genai.RoleUser
+		}
+		data := map[string]interface{}{
+			"id":             uuid.New().String(),
+			"span_id":        spanID,
```
cade-conklin, line 190: issue: `request_id` needs to be checked for from response headers

v3/integrations/nrgemini/nrgemini.go
```diff
@@ -0,0 +1,543 @@
...
+		"span_id":        spanID,
+		"trace_id":       traceID,
+		"role":           responseRole,
+		"completion_id":  completionID,
+		"sequence":       responseSeq,
+		"response.model": responseModel,
+		"vendor":         "gemini",
+		"ingest_source":  "Go",
+		"is_response":    true,
+	}
```
cade-conklin, line 230: issue: `request_id` needs to be checked for from response headers

v3/integrations/nrgemini/nrgemini.go
```diff
@@ -0,0 +1,543 @@
...
+	responseModel := model
+	if resp != nil && resp.ModelVersion != "" {
+		responseModel = resp.ModelVersion
+	}
+
+	for i, c := range contents {
+		text := extractContentText(c)
+		role := c.Role
+		if role == "" {
+			role = genai.RoleUser
```
cade-conklin, line 186: question: Why is this the default?

v3/integrations/nrgemini/nrgemini.go
```diff
@@ -0,0 +1,543 @@
...
+	if resp == nil || len(resp.Candidates) == 0 || resp.Candidates[0].Content == nil {
+		return ""
+	}
+	var parts []string
+	for _, p := range resp.Candidates[0].Content.Parts {
+		if p != nil && p.Text != "" {
+			parts = append(parts, p.Text)
+		}
+	}
+	return strings.Join(parts, " ")
```
cade-conklin, line 270: question: Is it less expensive to have `parts` be a string and just do a `parts = parts + " " + p.Text"`?

cade-conklin (changes requested):
So far only reviewed the simple example and have some requested changes + some questions.  Additionally, is this the only way that we can instrument chat messages? For instance, what about [this example](https://github.com/googleapis/go-genai/blob/main/examples/chats/chat.go) from the same package?

v3/integrations/nrgemini/nrgemini.go
```diff
@@ -0,0 +1,543 @@
...
+	if resp == nil {
+		return
+	}
+
+	responseText := extractResponseText(resp)
+	responseSeq := len(contents)
+	responseRole := genai.RoleModel
+	if len(resp.Candidates) > 0 && resp.Candidates[0].Content != nil && resp.Candidates[0].Content.Role != "" {
+		responseRole = resp.Candidates[0].Content.Role
+	}
```
cade-conklin, line 218: issue: `CandidateCount` is set in the `GenerateContentConfig` and can be greater than 1 (which is the default).  We only check the first index, so there could be more that we are missing

v3/integrations/nrgemini/nrgemini.go
```diff
@@ -0,0 +1,543 @@
...
+	return strings.Join(parts, " ")
+}
+
+func extractResponseText(resp *genai.GenerateContentResponse) string {
+	if resp == nil || len(resp.Candidates) == 0 || resp.Candidates[0].Content == nil {
+		return ""
+	}
+	var parts []string
+	for _, p := range resp.Candidates[0].Content.Parts {
```
cade-conklin, line 265: issue: Same thing here, what if there are multiple Candidates?

v3/integrations/nrgemini/nrgemini.go
```diff
@@ -0,0 +1,543 @@
...
+}
+
+func extractContentText(c *genai.Content) string {
+	if c == nil {
+		return ""
+	}
+	var parts []string
+	for _, p := range c.Parts {
+		if p != nil && p.Text != "" {
+			parts = append(parts, p.Text)
```
cade-conklin, line 254: suggestion: Add a else and a debug statement for content that is not captured

v3/integrations/nrgemini/nrgemini.go
cade-conklin, line 265: suggestion: Portions of `extractContentText` and `extractResponseText` are the same.  Pass in a parameter to a function to use the `range` keyword on and turn that into another function.

cade-conklin (changes requested):
Looking good so far.  Need a description before diving in to streaming - a high-level overview on the wrapper and its functions and what it accomplishes

mirackara:
Closing this PR out and splitting into two seperate PRs. One for AI event reporting fixes and one for gemini sdk.

https://github.com/newrelic/go-agent/pull/1209

https://github.com/newrelic/go-agent/pull/1208

